PR #22, opened 2016-09-09
adding Neo4jSparkSession.scala

Mageswaran1989:
Just an wrapper, inspired after Spark-2 SparkSession.

Review please.

Not yet ready for merge.

### DIFF
--- a/src/main/scala/org/neo4j/spark/Neo4jSparkSession.scala
+++ b/src/main/scala/org/neo4j/spark/Neo4jSparkSession.scala
@@ -1,7 +1,7 @@
 package org.neo4j.spark
 
 import org.apache.spark.graphx.Graph
-import org.apache.spark.sql.{types, SQLContext}
+import org.apache.spark.sql.{SparkSession, types, SQLContext}
 import org.apache.spark.{SparkConf, SparkContext}
 
 import scala.reflect.ClassTag
@@ -10,8 +10,8 @@ import scala.reflect.ClassTag
  * @author Mageswaran Dhandapani
  * @since 09.07.16
  */
-class Neo4jSparkSession private (sc: SparkContext,
-                                 sqlContext: SQLContext) {
+class Neo4jSparkSession private (@transient sc: SparkContext,
+                                 @transient sqlContext: SQLContext) {
 
   /**
    *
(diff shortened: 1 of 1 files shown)


src/main/scala/org/neo4j/spark/Neo4jSparkSession.scala
```diff
@@ -0,0 +1,184 @@
...
+    private[this] var userSpecifiedContext: Option[SparkContext] = None
+    private[this] var userSpecifiedSqlContext: Option[SQLContext] = None
+    private[this] var userSpecifiedSparkConf: Option[SparkConf] = None
+    private[this] var userSpecifiedBoltPassword: Option[String] = None
+    //private[this] var userSpecifiedSession: Option[SparkSession] = None //TODO SparkSession
+
+    val sparkContext = userSpecifiedContext.getOrElse {
+      val sparkConf =userSpecifiedSparkConf.getOrElse {
+        //User password is used when started inside an application
+        new SparkConf().set("spark.neo4j.bolt.password", userSpecifiedBoltPassword.getOrElse(""))
```
jexp: should it be getOrElseThrow ? or should that cover neo servers with auth disabled?

jexp:
hey, looks good, I have a similar thing going but not finished yet, I'll push it to a branch for review.

src/main/scala/org/neo4j/spark/Neo4jSparkSession.scala
Mageswaran1989: After testing, adding "throw" will enforce user from spark shell to use the boltPassword() call eventhough the config comes from the default spark config. What you think?

jexp:
I finished the loading API over the weekend and would love your feedback. 
Writing would be next.

See: https://github.com/neo4j-contrib/neo4j-spark-connector#builder-api

Perhaps we can align that with your Spark-Session like work? Happy to take suggestions / PR's on the new API.

The old APIs don't support partitions beyond one, so might not be continued.

Mageswaran1989:
wow! exactly that what I had in mind, but was not able put it on Scala as had some doubts on how to handle parameters across the different Spark components.

Let me take some time and get back to you

jexp:
@Mageswaran1989 were you able to have a look / test it out?

Mageswaran1989:
@jexp I had a look and completely carried away with something else...
please let me know your thoughts on how to proceed. I will catch up sometime today or tommorow

Mageswaran1989:
shall I patch your latest code with my changes for context /config?

jexp:
Perhaps quickly describe in a few words your intended changes / resoning, so that we can agree that these make sense. And then yes a PR would be welcome

Mageswaran1989:
https://github.com/neo4j-contrib/neo4j-spark-connector/blob/spark-20/src/main/scala/org/neo4j/spark/Neo4j.scala will be the file that is gonna be carried forwarded to Spark 2.x +, correct me if I wrong.

My intention was covered by you with above file changes. What I can do is to add a builder pattern to that like in Spark 2.

jexp:
Yes exactly this is the API that is going to be carried over, I just have to find the time to implement the write path.

What do you mean by:

> What I can do is to add a builder pattern to that like in Spark 2.

PR #33, opened 2017-03-13
Added builder pattern to retrive neo4j password

Mageswaran1989:
Example usage:

import org.neo4j.spark.Neo4j   
val neo4j = Neo4j.builder.sparkSession(spark).boltPassword("neo4j").getOrCreate()   
val rdd = neo4j.cypher("MATCH (n:Person) RETURN id(n) as id ").loadRowRdd   
rdd.count


Mageswaran1989:
Cclosing https://github.com/neo4j-contrib/neo4j-spark-connector/pull/22 since this PR covers it.

jexp:
Do we really need `.builder` ?

If we change this we also need to change the docs in the readme etc.

jexp:
We want to rework the APIs anyway, so we will switch to the Spark datasource API. 
Thank you for your contribution nonetheless.

Issue #120, opened 2020-01-10
Need driver update for Neo4j 4.0


moxious:
```
import org.neo4j.spark._
neo: org.neo4j.spark.Neo4j = org.neo4j.spark.Neo4j@5f492b15
beat: (String, Seq[String]) = (Beat,List(beat))
on_beat: (String, Seq[Nothing]) = (ON_BEAT,List())
crime: (String, Seq[String]) = (Crime,List(id, description, case_number, point))
crimeType: (String, Seq[String]) = (CrimeType,List(primary_type))
is_type: (String, Seq[Nothing]) = (IS_TYPE,List())
crimesDates: (String, Seq[String]) = (CrimeDate,List(parsed_date, date))
on_date: (String, Seq[Nothing]) = (ON_DATE,List())
org.apache.spark.SparkException: Job aborted due to stage failure: Task 0 in stage 5.0 failed 1 times, most recent failure: Lost task 0.0 in stage 5.0 (TID 5, localhost, executor driver): org.neo4j.driver.v1.exceptions.ClientException: The server does not support any of the protocol versions supported by this driver. Ensure that you are using driver and server versions that are compatible with one another.
	at org.neo4j.driver.internal.async.HandshakeHandler.protocolNoSupportedByServerError(HandshakeHandler.java:147)
	at org.neo4j.driver.internal.async.HandshakeHandler.decode(HandshakeHandler.java:123)
	at org.neo4j.driver.internal.shaded.io.netty.handler.codec.ByteToMessageDecoder.decodeRemovalReentryProtection(ByteToMessageDecoder.java:489)
	at org.neo4j.driver.internal.shaded.io.netty.handler.codec.ReplayingDecoder.callDecode(ReplayingDecoder.java:367)
	at org.neo4j.driver.internal.shaded.io.netty.handler.codec.ByteToMessageDecoder.channelRead(ByteToMessageDecoder.java:265)
	at org.neo4j.driver.internal.shaded.io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:362)
	at org.neo4j.driver.internal.shaded.io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:348)
	at org.neo4j.driver.internal.shaded.io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:340)
[...]

PR #121, opened 2020-02-06
fixes: #120: Need driver update for Neo4j 4.0

conker84:
List of tasks:
- [x] updated driver to Neo4j v 4.0
- [x] added support to multi-tenancy via option `spark.neo4j.database`
- [x] improved tests with Testcontainers


pom.xml
```diff
@@ -39,15 +39,27 @@
     <properties>
         <project.scala.version>2.11.12</project.scala.version>
         <project.scala.binary.version>2.11</project.scala.binary.version>
-
-        <neo4j.version>3.5.12</neo4j.version>
-        <driver.version>1.7.5</driver.version>
-        <spark.version>2.4.0</spark.version>
+        <neo4j.version>4.0.0</neo4j.version>
+        <driver.version>4.0.0</driver.version>
+        <spark.version>2.4.1</spark.version>
```
krisgeus: Spark 2.4.4 is the latest and greatest spark version out there. Might be good to use that one.

krisgeus: Other than that LGTM

src/main/scala/org/neo4j/spark/rdd/Neo4jRowRDD.scala
```diff
@@ -20,31 +20,31 @@ class Neo4jRowRDD(@transient sc: SparkContext, val query: String, val parameters
...
+                i = i + 1
+              }
+              Row.fromSeq(builder.result())
+            }
+            res
+          })
+        }.iterator
+      }
+      session.readTransaction(runner)
     } finally {
```
moxious: I think this needs experimentation.   I could be wrong about this so we don't necessarily need a change but when I was playing with my own private branch adapting spark for 4.0, these "finally close the session" blocks were a big problem.   This needs more testing.

Reason is that under the new reactive driver setup, the semantics of how/when records are consumed has changed.   Spark itself is so lazy, when you exit this block the very first record hasn't been pulled yet, but you're closing the session.   In my experimentation this broke badly because basically you'd always fail to get data back from neo4j saying that the driver or the session was closed.

For unit tests on this you might need some delay, and record sets with > 1 coming back.   Or you can create a test build, play with it in zeppelin

I'll not put the same comment on every finally block, but this pattern (run the query, return, close the session/driver in a finally block) is common in the code base.

moxious: See "important note on semantics" in the new java driver API docs -- I think this interacting with spark's laziness is specifically problematic.  https://neo4j.com/docs/api/java-driver/4.0/org/neo4j/driver/QueryRunner.html

conker84: @moxious thanks a lot for the feedback. We definitely need to improve this, I have in mind to do it with a new release that supports Spark Datasource APIs.

About transaction management, looking at the code I invoke `Result#list` which store the whole result for later use
https://neo4j.com/docs/api/java-driver/4.0/org/neo4j/driver/Result.html#list--

This is the only way that I found to interact with the db that allows us to use transaction functions (that should allow us to work properly with Aura)

Wdyt?

src/main/scala/org/neo4j/spark/Executor.scala
```diff
@@ -35,79 +35,50 @@ object Executor {
...
+      val keys = peek.keys().asScala
+      val fields = keys.map(k => (k, peek.get(k).`type`())).map(keyType => CypherTypes.field(keyType))
+      val schema = StructType(fields)
+
+      val it = result.asScala.map(record => {
+        val row = new Array[Any](keyCount)
+        var i = 0
+        while (i < keyCount) {
+          val value = record.get(i).asObject() match {
+            case it: util.Map[_, _] => it.asScala.toString()
```
jexp: why toString?

src/main/scala/org/neo4j/spark/Executor.scala
```diff
@@ -35,79 +35,50 @@ object Executor {
...
-            while (i < keyCount) {
-              val value = record.get(i).asObject() match {
-                case it: util.Map[_, _] => it.asScala
-                case it: util.Collection[_] => it.toArray()
-                case x => x
-              }
-              row.update(i, value)
-              i = i + 1
-            }
-            if (!result.hasNext) {
```
jexp: how do we handle closing of resources now?

src/main/scala/org/neo4j/spark/Executor.scala
```diff
@@ -35,79 +35,50 @@ object Executor {
...
-            }
-            row
-          })
-          new CypherResult(schema, it)
+          row.update(i, value)
+          i = i + 1
         }
+        row
+      })
+      new CypherResult(schema, new LazySessionIterator(it, driver, session))
```
jexp: esp. how does it work to return/leak the driver+session+result out of the spark job processing?

src/main/scala/org/neo4j/spark/Executor.scala
```diff
@@ -35,79 +35,50 @@ object Executor {
...
+    } catch {
+      case e: Exception => {
+        if (session.isOpen) {
```
jexp: null checks?

src/main/scala/org/neo4j/spark/Executor.scala
```diff
@@ -35,79 +35,50 @@ object Executor {
...
-        }
-        driver.close()
-      } catch {
-        case _ => // ignore
-      }
-    }
-
     val driver: Driver = config.driver()
-    val session = driver.session()
+    val session = driver.session(config.sessionConfig())
```
jexp: what closes the session and driver? shouldn't be there a finally or such?

src/main/scala/org/neo4j/spark/Executor.scala
```diff
@@ -35,79 +35,50 @@ object Executor {
...
+    } catch {
+      case e: Exception => {
+        if (session.isOpen) {
+          session.close()
+        }
+        driver.close()
+        throw e
       }
-
-      if (write)
```
jexp: the read and write transactions are very important for routing in a causal cluster so we must not lose that functionality

src/main/scala/org/neo4j/spark/Executor.scala
```diff
@@ -35,79 +35,50 @@ object Executor {
...
-            return new CypherResult(new StructType(), Iterator.empty)
-          }
-          val peek = result.peek()
-          val keyCount = peek.size()
-          if (keyCount == 0) {
-            val res: CypherResult = new CypherResult(new StructType(), Array.fill[Array[Any]](rows(result))(EMPTY).toIterator)
-            result.consume()
-            close(driver, session)
-            return res
+      val result: Result = session.run(query, toJava(parameters))
```
jexp: session.run doesn't cut it, see below, we need to use tx-functions for routing and retries

src/main/scala/org/neo4j/spark/LazySessionIterator.scala
```diff
@@ -0,0 +1,32 @@
+package org.neo4j.spark
+
+import org.neo4j.driver.Driver
+import org.neo4j.driver.Session
+
+class LazySessionIterator[T](iterator: Iterator[T], driver: Driver, session: Session) extends Iterator[T] {
```
jexp: I'm not suer this really works ... can this leak out of scope of the worker job so that the driver/session might not be closed (if no one calls close here) there is no guarantee or?

src/main/scala/org/neo4j/spark/LazySessionIterator.scala
```diff
@@ -0,0 +1,32 @@
...
+  private lazy val it = iterator
+
+  override def hasNext: Boolean = {
+    val hasNext = it.hasNext
```
jexp: also has to close if if it.next throws an error

src/main/scala/org/neo4j/spark/LazySessionIterator.scala
```diff
@@ -0,0 +1,32 @@
...
+    if (!hasNext) {
+      close
+    }
+    hasNext
+  }
+
+  private def close = {
+    if (session.isOpen) session.close()
```
jexp: null check on session and wrap with try-catch in case session.close throws  an exception

src/main/scala/org/neo4j/spark/Neo4jConfig.scala
```diff
@@ -1,34 +1,41 @@
...
 import org.apache.spark.SparkConf
-import org.neo4j.driver.v1.{Driver, AuthTokens, Config, GraphDatabase}
+import org.neo4j.driver.{AuthToken, AuthTokens, Config, Driver, GraphDatabase, Session, SessionConfig}
 
 /**
  * @author mh
  * @since 02.03.16
  */
-case class Neo4jConfig(val url: String, val user: String = "neo4j", val password: Option[String] = None, encryptionStatus:Boolean) {
+case class Neo4jConfig(val url: String,
```
jexp: default value for encryption (not sure where "status" comes from there)

src/main/scala/org/neo4j/spark/Neo4jConfig.scala
```diff
@@ -1,34 +1,41 @@
...
 
 import org.apache.spark.SparkConf
-import org.neo4j.driver.v1.{Driver, AuthTokens, Config, GraphDatabase}
+import org.neo4j.driver.{AuthToken, AuthTokens, Config, Driver, GraphDatabase, Session, SessionConfig}
 
 /**
  * @author mh
  * @since 02.03.16
  */
-case class Neo4jConfig(val url: String, val user: String = "neo4j", val password: Option[String] = None, encryptionStatus:Boolean) {
```
jexp: should we also add support for multi-db here? specify the db to connect to?

src/main/scala/org/neo4j/spark/Neo4jConfig.scala
```diff
@@ -1,34 +1,41 @@
...
   def apply(sparkConf: SparkConf): Neo4jConfig = {
-    val url = sparkConf.get(prefix + "url", "bolt://localhost")
-    val user = sparkConf.get(prefix + "user", "neo4j")
-    val password: Option[String] = sparkConf.getOption(prefix + "password")
-    val encryptionStatus : Boolean = sparkConf.getBoolean(prefix + "encryption.status", defaultValue = false)
-    Neo4jConfig(url, user, password,encryptionStatus)
+    val url = sparkConf.get(s"$prefix.bolt.url", "bolt://localhost")
+    val user = sparkConf.get(s"$prefix.bolt.user", "neo4j")
+    val password: Option[String] = sparkConf.getOption(s"$prefix.bolt.password")
+    val database: Option[String] = sparkConf.getOption(s"$prefix.database")
```
jexp: why no bolt here ? seems inconsistent?

src/main/scala/org/neo4j/spark/Neo4jConfig.scala
```diff
@@ -1,34 +1,41 @@
...
+    val encryptionStatus : Boolean = sparkConf.getBoolean(s"$prefix.bolt.encryption.status", defaultValue = false)
```
jexp: no need for ".status"

src/main/scala/org/neo4j/spark/dataframe/Neo4jDataFrame.scala
```diff
@@ -206,38 +145,35 @@ object Neo4jDataFrame {
 
     override def compute(split: Partition, context: TaskContext): Iterator[Row] = {
       val driver = config.driver()
-      val session = driver.session()
+      val session = driver.session(config.sessionConfig())
+      val result = session.run(query, params)
```
jexp: sorry, doesn't work, we need the transaction function for routing and retries

src/main/scala/org/neo4j/spark/rdd/Neo4jRowRDD.scala
```diff
@@ -20,12 +20,11 @@ class Neo4jRowRDD(@transient sc: SparkContext, val query: String, val parameters
...
   override def compute(split: Partition, context: TaskContext): Iterator[Row] = {
     val driver = config.driver()
-    val session = driver.session()
+    val session = driver.session(config.sessionConfig())
 
     try {
-      val result: StatementResult = session.run(query, parameters.toMap.mapValues(_.asInstanceOf[AnyRef]).asJava)
-
-      result.asScala.map(record => {
+      val result = session.run(query, parameters.toMap.mapValues(_.asInstanceOf[AnyRef]).asJava)
```
jexp: same here

src/main/scala/org/neo4j/spark/rdd/Neo4jRowRDD.scala
```diff
@@ -39,15 +38,17 @@ class Neo4jRowRDD(@transient sc: SparkContext, val query: String, val parameters
...
-        }
         res
       })
-    } finally {
-      if (session.isOpen) session.close()
-      driver.close()
+      new LazySessionIterator(it, driver, session)
+    } catch {
+      case e: Exception => {
+        if (session.isOpen) {
```
jexp: null checks and try-catch for session closing 
that's why I wrapped this in a function that I could reuse in all the places !

src/main/scala/org/neo4j/spark/rdd/Neo4jTupleRDD.scala
```diff
@@ -14,21 +19,21 @@ class Neo4jTupleRDD(@transient sc: SparkContext, val query: String, val paramete
...
-      val result = session.run(query, parameters.toMap.asJava)
-
-      result.asScala.map((record) => {
-        val res = record.asMap().asScala.toSeq
-        if (!result.hasNext) {
-          if (session.isOpen) session.close()
-          driver.close()
-        }
-        res
+      val result: Result = session.run(query, parameters.toMap.asJava)
```
jexp: same here with tx function

src/main/scala/org/neo4j/spark/rdd/Neo4jTupleRDD.scala
```diff
@@ -14,21 +19,21 @@ class Neo4jTupleRDD(@transient sc: SparkContext, val query: String, val paramete
...
+      val it = result.asScala.map((record) => {
+        record.asMap().asScala.toSeq
       })
-    } finally {
-      if (session.isOpen) session.close()
-      driver.close()
+      new LazySessionIterator(it, driver, session)
+    } catch {
+      case e: Exception => {
+        if (session.isOpen) {
```
jexp: same here

src/test/scala/org/neo4j/spark/SparkConnectorScalaSuiteIT.scala
```diff
@@ -0,0 +1,48 @@
...
+
+import org.apache.spark.{SparkConf, SparkContext}
+import org.junit.{AfterClass, Assume, BeforeClass}
+import org.junit.runner.RunWith
+import org.junit.runners.Suite
+import org.neo4j.Neo4jContainerExtension
+
+
+object SparkConnectorScalaSuiteIT {
+  val server: Neo4jContainerExtension = new Neo4jContainerExtension("neo4j:4.0.0-enterprise")
```
jexp: should we use latest-enterprise?

src/test/scala/org/neo4j/spark/SparkConnectorScalaSuiteIT.scala
```diff
@@ -0,0 +1,48 @@
...
+  @BeforeClass
+  def setUpContainer() = {
+    if (!server.isRunning) {
+      try {
+        server.start()
+      } catch {
+        case _ => //
+      }
+      Assume.assumeTrue("Neo4j container is not started", server.isRunning)
+      conf = new SparkConf().setAppName("neoTest")
```
jexp: Should we set up spark with testcontainers too?

src/main/scala/org/neo4j/spark/Neo4jConfig.scala
conker84: There is the multi support db via config property `spark.neo4j.database`.

src/main/scala/org/neo4j/spark/Neo4jConfig.scala
conker84: Because it's not bolt strictly related to I used `spark.neo4j.database`
Do you prefer `spark.neo4j.bolt.database`?

src/main/scala/org/neo4j/spark/Neo4jConfig.scala
conker84: done

src/test/scala/org/neo4j/spark/SparkConnectorScalaSuiteIT.scala
conker84: done

src/main/scala/org/neo4j/spark/rdd/Neo4jRowRDD.scala
conker84: ok

src/test/scala/org/neo4j/spark/SparkConnectorScalaSuiteIT.scala
conker84: Yes absolutely, can we do it in a separate PR?

src/main/scala/org/neo4j/spark/Executor.scala
conker84: restored, I realized that the test `runCypherQueryWithSchemaAndMap` was wrong, I fixed it

README.md
```diff
@@ -69,7 +69,7 @@ Otherwise set the `spark.neo4j.bolt.url` in your `SparkConf` pointing e.g. to `b
 You can provide user and password as part of the URL `bolt://neo4j:<password>@localhost` or individually in `spark.neo4j.bolt.user` and `spark.neo4j.bolt.password`.
 
 If you're running Neo4j with the Bolt connector and the option `dbms.connector.bolt.tls_level` in Neo4j is `REQUIRED`, you must set the `spark.neo4j.bolt.encryption.status` to `true` in your `SparkConf`.
-Otherwise, you can either ignore `spark.neo4j.bolt.encryption.status` or  set  `spark.neo4j.bolt.encryption.status` to `false` in your `SparkConf`.
+Otherwise, you can either ignore `spark.neo4j.bolt.encryption` or  set  `spark.neo4j.bolt.encryption` to `false` in your `SparkConf`.
 
 ## Builder API
```
jexp: can you make sure in the readme the other version numbers that are mentioned are fixed / correct?

src/main/scala/org/neo4j/spark/Executor.scala
jexp: ping?

src/main/scala/org/neo4j/spark/Executor.scala
jexp: ping?

src/main/scala/org/neo4j/spark/Neo4jConfig.scala
jexp: either we use bolt everywhere or we drop it everywhere (then we have to update docs + readme)
user and password are also not related to bolt :)

moxious: Driver changes in 4.0 have sadly made this one complicated.   bolt will continue to work, but it's deprecated, and implies the wrong thing for 3.5 cluster.   Neo4j is new hotness -- but *does not* work with 3.5.

My vote would be to make it `spark.neo4j.database` and use "neo4j" throughout to be supportive of the overall messaging for Neo4j.  But 100% we're going to need some special instructions for migration of the old driver, and an explanation of all of this bolt, bolt+routing, and neo4j craziness.

conker84: @jexp you're right I'm removing bolt from everywhere and updating the docs.

src/main/scala/org/neo4j/spark/dataframe/Neo4jDataFrame.scala
```diff
@@ -21,7 +22,9 @@ object Neo4jDataFrame {
                     source: (String, Seq[String]),
                     relationship: (String, Seq[String]),
                     target: (String, Seq[String]),
-                    renamedColumns: Map[String, String] = Map.empty): Unit = {
+                    renamedColumns: Map[String, String] = Map.empty,
+                    partitions: Int = 1,
+                    unwindBatchSize: Int = 10000): Unit = {
     dataFrame.cache()
```
jexp: should this by default? isn't that something that the user wants to control?

src/main/scala/org/neo4j/spark/dataframe/Neo4jDataFrame.scala
```diff
@@ -32,56 +35,66 @@ object Neo4jDataFrame {
         MERGE (target:${target._1.quote} {${targetLabel.quote} : row.target.${targetLabel.quote}}) ON CREATE SET target += row.target
```
jexp: we should split this up, into two statements like in saveGraph

src/main/scala/org/neo4j/spark/dataframe/Neo4jDataFrame.scala
```diff
@@ -32,56 +35,66 @@ object Neo4jDataFrame {
...
+                  partitions: Int = 1,
+                  unwindBatchSize: Int = 10000): Unit = {
     dataFrame.cache()
     val nodeLabel: String = renamedColumns.getOrElse(nodes._2.head, nodes._2.head)
     val createStatement =
       s"""
         UNWIND $$rows as row
-        CREATE (node:${nodes._1.quote} {${nodeLabel.quote} : row.source.${nodeLabel.quote}})
-        SET node = row.node_properties
+        CREATE (node:${nodes._1.quote} {${nodeLabel.quote} : row.node_properties.${nodeLabel.quote}})
```
jexp: possibly make this configurable via parameter (GH issue)

jexp: MERGE vs. CREATE I mean

src/main/scala/org/neo4j/spark/dataframe/Neo4jDataFrame.scala
```diff
@@ -32,56 +35,66 @@ object Neo4jDataFrame {
...
-      }
-      if (write) {
-        session.writeTransaction(runner)
-      }
-      else
-        session.readTransaction(runner)
-    } finally {
-      if (session.isOpen) session.close()
-      driver.close()
+  def execute(session: Session, query: String, parameters: java.util.Map[String, AnyRef], write: Boolean = false): ResultSummary = {
```
jexp: should this be private?

src/main/scala/org/neo4j/spark/dataframe/Neo4jDataFrame.scala
```diff
@@ -111,88 +124,26 @@ object Neo4jDataFrame {
...
-//        }
-//        res
-//      })
-//    }
-//
-//    override protected def getPartitions: Array[Partition] = Array(new Neo4jPartition())
-//  }
-
   def apply(sqlContext: SQLContext, query: String, parameters: java.util.Map[String, AnyRef]): DataFrame = {
     val limitedQuery = s"$query LIMIT 1"
```
jexp: don't remember where this was from ...

src/main/scala/org/neo4j/spark/dataframe/Neo4jDataFrame.scala
```diff
@@ -206,13 +157,13 @@ object Neo4jDataFrame {
...
-      val session = driver.session()
+      val session = driver.session(config.sessionConfig())
       try {
-        val runTransaction = new TransactionWork[Iterator[Row]]() {
+        val txWork = new TransactionWork[Iterator[Row]] {
           override def execute(tx: Transaction): Iterator[Row] = {
             val result = tx.run(query, params)
             if (!result.hasNext) throw new RuntimeException("Can't determine schema from empty result")
-            result.asScala.map(record => {
+            result.list().asScala.map(record => {
```
jexp: this is potentially a lot that's eagerly pulled ... (think 10 or 100M or 1bn rows ...) ?
I meant to defer the materialization as the iterator is used by spark
and close resources after the last record has been consumed.

src/test/scala/org/neo4j/spark/SparkConnectorScalaBaseTSE.scala
```diff
@@ -0,0 +1,49 @@
...
+    if (driver == null) {
+      driver = GraphDatabase.driver(SparkConnectorScalaSuiteIT.server.getBoltUrl, AuthTokens.none())
+    }
+    driver.session
+  }
+
+  @Before
+  @throws[Exception]
+  def start() {
+    session().run("MATCH (n) DETACH DELETE n").consume()
```
jexp: for large tests this might exceed memory

src/test/scala/org/neo4j/spark/SparkConnectorScalaBaseTSE.scala
```diff
@@ -0,0 +1,49 @@
...
+
+class SparkConnectorScalaBaseTSE {
+
+  val conf: SparkConf = SparkConnectorScalaSuiteIT.conf
+  val sc: SparkContext = SparkConnectorScalaSuiteIT.sc
+
+  private var driver: Driver = _
+
+  def session(): Session = {
+    if (driver == null) {
```
jexp: shouldn't we allocate teh driver at the beginning of the test?

src/main/scala/org/neo4j/spark/utils/Neo4jUtils.scala
```diff
@@ -0,0 +1,16 @@
+package org.neo4j.spark.utils
+
+import org.neo4j.driver.{Driver, Session}
+
+object Neo4jUtils {
+
+  def close(driver: Driver, session: Session): Unit = {
+    if (session != null && session.isOpen) {
```
jexp: wrap with try-catch to guarantee to close the driver even if closing the session fails for some reason,
want to avoid leaks

src/main/scala/org/neo4j/spark/rdd/Neo4jTupleRDD.scala
```diff
@@ -14,21 +15,19 @@ class Neo4jTupleRDD(@transient sc: SparkContext, val query: String, val paramete
...
-
-      result.asScala.map((record) => {
-        val res = record.asMap().asScala.toSeq
-        if (!result.hasNext) {
-          if (session.isOpen) session.close()
-          driver.close()
+      val txWork = new TransactionWork[Iterator[Seq[(String, AnyRef)]]] {
+        override def execute(tx: Transaction): Iterator[Seq[(String, AnyRef)]] = {
+          val result: Result = tx.run(query, parameters.toMap.asJava)
+          result.list().asScala.map((record) => {
```
jexp: same as in the other place, i'm a bit afraid of materializing all results eagerly and not as spark consumes the iterator ...

README.md
```diff
@@ -255,22 +255,25 @@ There are a few different RDD's all named `Neo4jXxxRDD`
...
     * the result are sent in batches of 10000 to the graph
-    * optional `renamedColumns` parameter - can be used to create a relationship between nodes having the same properties.<br/>
+    * optional `renamedColumns` parameter - can be used to create a relationship between nodes having the same properties.
+    * the `partitions` parameter defines the number of partitions of the dataframe will be repartitioned into. It defines the level of the parallelism, so consider that it could lead to transaction locks if the same node gets managed concurrently;
+    * the `unwindBatchSize` parameter defines for each partition the batch size sent to Neo4j.<br/>
  *For example*: `(keanu:Person {name: 'Keanu'})-[:ACTED_WITH]->(Laurence:Person {name: "Laurence"})` requires a DataFrame with 2 `name` columns which is not possible.
```
jexp: the example is for renamedColums, need to keep it up there.

src/main/scala/org/neo4j/spark/rdd/Neo4jRowRDD.scala
```diff
@@ -20,34 +21,32 @@ class Neo4jRowRDD(@transient sc: SparkContext, val query: String, val parameters
...
-          }
-          Row.fromSeq(builder.result())
-        }
-        if (!result.hasNext) {
-          if (session.isOpen) session.close()
-          driver.close()
+      val txWork = new TransactionWork[Iterator[Row]] {
+        override def execute(tx: Transaction): Iterator[Row] = {
+          val result = tx.run(query, parameters.toMap.mapValues(_.asInstanceOf[AnyRef]).asJava)
+          result.list().asScala.map(record => {
```
jexp: same here ...

src/main/scala/org/neo4j/spark/rdd/Neo4jRowRDD.scala
```diff
@@ -20,34 +21,32 @@ class Neo4jRowRDD(@transient sc: SparkContext, val query: String, val parameters
...
+            val keyCount = record.size()
+
+            val res = if (keyCount == 0) Row.empty
+            else if (keyCount == 1) Row(convert(record.get(0).asObject()))
+            else {
+              val builder = Seq.newBuilder[AnyRef]
```
jexp: seed with keyCount

src/main/scala/org/neo4j/spark/rdd/Neo4jRowRDD.scala
jexp: it's definitely a tricky area. I think we should discuss with the drivers team what's the best interaction pattern that gives us all

- lazyness / reactive
- transaction functions
- ensure closing driver / session to avoid leaks
- support sparks lazy evaluation
- avoid pre-materialization of results

src/test/scala/org/neo4j/spark/Neo4jGraphScalaTSE.scala
```diff
@@ -0,0 +1,86 @@
...
+  */
+class Neo4jGraphScalaTSE extends SparkConnectorScalaBaseTSE {
+  val FIXTURE: String = "CREATE (s:A {a:0})-[r:REL {foo:'bar'}]->(t:B {b:1}) RETURN id(s) AS source, id(t) AS target"
+
+  private var source: Long = _
+  private var target: Long = _
+
+  @Before
+  @throws[Exception]
+  def setUp {
```
jexp: should we do these tests with larger numbers that exceed partition sizes?

src/main/scala/org/neo4j/spark/rdd/Neo4jRowRDD.scala
conker84: can we organize a chat with them?

moxious: I'll ask if we can put together that chat.  I don't have an opinion on a solution approach because this stuff is too new to me and I don't understand the consequences either way.

README.md
conker84: Done

src/main/scala/org/neo4j/spark/dataframe/Neo4jDataFrame.scala
conker84: ok

src/main/scala/org/neo4j/spark/dataframe/Neo4jDataFrame.scala
conker84: I added it in this PR

src/main/scala/org/neo4j/spark/dataframe/Neo4jDataFrame.scala
conker84: done

src/test/scala/org/neo4j/spark/SparkConnectorScalaBaseTSE.scala
conker84: Yes I agree with you, I did that in order to start the Neo4j container just once, for large tests  we should think about a different strategy.

src/test/scala/org/neo4j/spark/SparkConnectorScalaBaseTSE.scala
conker84: The container instance is one for all tests so I think we can do it here

src/main/scala/org/neo4j/spark/utils/Neo4jUtils.scala
conker84: perfect, done

README.md
conker84: done

src/main/scala/org/neo4j/spark/rdd/Neo4jRowRDD.scala
conker84: done

src/test/scala/org/neo4j/spark/SparkConnectorScalaSuiteIT.scala
jexp: sure

src/main/scala/org/neo4j/spark/dataframe/Neo4jDataFrame.scala
jexp: ping?

src/main/scala/org/neo4j/spark/rdd/Neo4jTupleRDD.scala
jexp: ping @conker84 ?

src/test/scala/org/neo4j/spark/Neo4jGraphScalaTSE.scala
jexp: ping @conker84

src/main/scala/org/neo4j/spark/Executor.scala
moxious: @jexp @conker84 Nigel on the drivers team said that the API still provides for the old style sessions which are not lazy/reactive, and we should use those

src/test/scala/org/neo4j/spark/Neo4jGraphScalaTSE.scala
conker84: Yes I'll do that in a separate issue ok? I created #123

conker84:
> Some of my questions din't have a feedback can you have another look.
> Could you summarize the changes you did after my last feedback esp. around the tx functions result.list resource (close) handling etc?

I answered to the question as I found the solution, I'm still working on the others (tx-fun and resource handling) I'll be back once I found a solution

jexp:
@conker84 did you test the changes with some larger (e.g. 10M nodes) read and write setups? (with multiple partitions)
And with the aura instance too?

conker84:
@jexp as we discussed on Slack I implemented the Unmanaged Transaction flow.
I added a test `after` each method that checks if there are leaked transaction in that test execution.
I get the current active sessions at the container startup via the following code:

https://github.com/neo4j-contrib/neo4j-spark-connector/blob/1e677e7d8266716abe0fc33b5318f29f080f955a/src/test/scala/org/neo4j/spark/SparkConnectorScalaSuiteIT.scala#L66-L76

And here:

https://github.com/neo4j-contrib/neo4j-spark-connector/blob/1e677e7d8266716abe0fc33b5318f29f080f955a/src/test/scala/org/neo4j/spark/SparkConnectorScalaSuiteIT.scala#L45

Then in the after phase of each test I do the following check:

https://github.com/neo4j-contrib/neo4j-spark-connector/blob/1e677e7d8266716abe0fc33b5318f29f080f955a/src/test/scala/org/neo4j/spark/SparkConnectorScalaBaseTSE.scala#L48-L53

I tried to apply al your feedback if I lost a comment please apologise in advance.

I also tried with cloud-to-cloud import from Databricks CE to AURA (the smallest instance),
I was able to import a csv of 2 milion rows (the max size without getting an OOM from Databricks CE) in around 44 mins. Please check the following image:

Schermata 2020-03-30 alle 22 52 37

README.md
```diff
@@ -62,17 +62,18 @@ In your pom.xml, add:
 
 ## Config
 
-If you're running Neo4j on localhost with the default ports, you only have to configure your password in `spark.neo4j.bolt.password=<password>`.
```
jexp, line 65: can we highlight prominently that the config has changed?
do we still support the old config as a fallback?

pom.xml
```diff
@@ -39,15 +39,28 @@
     <properties>
         <project.scala.version>2.11.12</project.scala.version>
         <project.scala.binary.version>2.11</project.scala.binary.version>
-
-        <neo4j.version>3.5.12</neo4j.version>
-        <driver.version>1.7.5</driver.version>
-        <spark.version>2.4.0</spark.version>
+        <neo4j.version>4.0.1</neo4j.version>
+        <driver.version>4.0.1</driver.version>
```
jexp, line 43: 4.0.2 ?

pom.xml
```diff
@@ -39,15 +39,28 @@
     <properties>
         <project.scala.version>2.11.12</project.scala.version>
         <project.scala.binary.version>2.11</project.scala.binary.version>
-
-        <neo4j.version>3.5.12</neo4j.version>
-        <driver.version>1.7.5</driver.version>
-        <spark.version>2.4.0</spark.version>
+        <neo4j.version>4.0.1</neo4j.version>
```
jexp, line 42: 4.0.3

src/main/scala/org/neo4j/spark/Executor.scala
```diff
@@ -1,32 +1,19 @@
...
-import org.neo4j.driver.v1.{Driver, Session, StatementResult, Transaction, TransactionWork}
 import org.neo4j.spark.dataframe.CypherTypes
+import org.neo4j.spark.utils.Neo4jSessionAwareIterator
 
 import scala.collection.JavaConverters._
 
 
 object Executor {
 
-  def convert(value: AnyRef): Any = value match {
```
jexp, line 20: what happend with this?

src/main/scala/org/neo4j/spark/Executor.scala
```diff
@@ -48,75 +37,39 @@ object Executor {
...
+      return new CypherResult(new StructType(), Array.fill[Array[Any]](rows(result))(EMPTY).toIterator)
+    }
+    val keys = peek.keys().asScala
+    val fields = keys.map(k => (k, peek.get(k).`type`())).map(keyType => CypherTypes.field(keyType))
+    val schema = StructType(fields)
+    val it = result.map(record => {
+      val row = new Array[Any](keyCount)
+      var i = 0
+      while (i < keyCount) {
+        val value = record.get(i).asObject() match {
```
jexp, line 63: datetime conversion was removed!! please re-add

src/main/scala/org/neo4j/spark/Executor.scala
```diff
@@ -48,75 +37,39 @@ object Executor {
...
-          }
-          val keys = peek.keys().asScala
-          val fields = keys.map(k => (k, peek.get(k).`type`())).map(keyType => CypherTypes.field(keyType))
-          val schema = StructType(fields)
-
-          val it = result.asScala.map(record => {
-            val row = new Array[Any](keyCount)
-            var i = 0
-            while (i < keyCount) {
-              val value = convert(record.get(i).asObject())
```
jexp, line 99: pleas re-add this bit

src/main/scala/org/neo4j/spark/Neo4jConfig.scala
```diff
@@ -1,34 +1,44 @@
...
+
+  def sessionConfig(write: Boolean = false): SessionConfig = database.map { SessionConfig.builder().withDatabase(_) }
+    .getOrElse(SessionConfig.builder())
+    .withDefaultAccessMode(if (write) AccessMode.WRITE else AccessMode.READ)
+    .build()
 
 }
 
 object Neo4jConfig {
-  val prefix = "spark.neo4j.bolt."
```
jexp, line 26: should we as a fallback support the old syntax for a version or two?

README.md
conker84: Should we support the old config?

pom.xml
conker84: ok

pom.xml
conker84: ok

src/main/scala/org/neo4j/spark/Executor.scala
conker84: I'm sorry something went wrong when I rebased the code.
I'll re-add this and others part

src/main/scala/org/neo4j/spark/Executor.scala
conker84: I'm sorry something went wrong when I rebased the code.
I'll re-add this and others part

src/main/scala/org/neo4j/spark/Executor.scala
conker84: I'm sorry something went wrong when I rebased the code.
I'll re-add this and others part

src/main/scala/org/neo4j/spark/Neo4jConfig.scala
conker84: Oki

pom.xml
conker84: The last one on maven central is:
https://mvnrepository.com/artifact/org.neo4j.driver/neo4j-java-driver

conker84:
@jexp I updated the PR thanks a lot for the feedback!

pom.xml
jexp: oh sorry

README.md
conker84: I added this: `**Nb.**  The old spark config prefix `spark.neo4j.bolt` is deprecated and will be removed in the next release.`

conker84:
@jexp I'm investigating why is falling on Travis while in my local env the build is successful

conker84:
@jexp it's green now

conker84:
I wrapped the connection check with and `assertEventually`:

https://github.com/neo4j-contrib/neo4j-spark-connector/blob/37a69a39faa5fe972ef86cea5514167231eaf717/src/test/scala/org/neo4j/spark/SparkConnectorScalaBaseTSE.scala#L55-L61

src/main/scala/org/neo4j/spark/dataframe/Neo4jDataFrame.scala
```diff
@@ -24,67 +23,78 @@ object Neo4jDataFrame {
...
-        session.readTransaction(runner)
-    } finally {
-      if (session.isOpen) session.close()
-      driver.close()
+  private def execute(session: Session, query: String, parameters: java.util.Map[String, AnyRef], write: Boolean = false): ResultSummary = {
+    val runner = new TransactionWork[ResultSummary]() {
+      override def execute(tx: Transaction): ResultSummary =
+        tx.run(query, parameters).consume()
+    }
+    if (write) {
```
jexp, line 94: do we still need all this incl. the TransactionWork if we have the defaultTxMode?

src/main/scala/org/neo4j/spark/dataframe/Neo4jDataFrame.scala
jexp: a just to determine the datatypes

src/main/scala/org/neo4j/spark/dataframe/Neo4jDataFrame.scala
```diff
@@ -265,7 +199,7 @@ object CypherTypes {
     case "BOOLEAN" => BOOLEAN
     case "BOOL" => BOOLEAN
     case "DATETIME" => DATETIME
-	case "TIME" => TIME
+    case "TIME" => DATETIME
```
jexp, line 202: Are you sure?

jexp: TIME -> DATETIME?

src/main/scala/org/neo4j/spark/rdd/Neo4jTupleRDD.scala
```diff
@@ -13,23 +16,9 @@ class Neo4jTupleRDD(@transient sc: SparkContext, val query: String, val paramete
...
-          driver.close()
-        }
-        res
-      })
-    } finally {
-      if (session.isOpen) session.close()
-      driver.close()
-    }
+    new Neo4jSessionAwareIterator(config, query, parameters.toMap.asJava, false).map(record => {
+      record.asMap().asScala.toSeq
```
jexp, line 20: do we have to convert to Spark types here? Perhaps separate PR/issue

src/main/scala/org/neo4j/spark/utils/Neo4jSessionAwareIterator.scala
```diff
@@ -0,0 +1,62 @@
...
+
+  override def hasNext: Boolean = {
+    try {
+      val hasNext = result.hasNext
+      if (!hasNext) {
+        close()
+      }
+      hasNext
+    } catch {
+      case _ => {
```
jexp, line 27: Do we plan to report errors somewhere?

src/main/scala/org/neo4j/spark/utils/Neo4jSessionAwareIterator.scala
```diff
@@ -0,0 +1,62 @@
...
+        if (rollback && write) {
+          transaction.rollback()
+        } else {
+          transaction.commit()
+        }
+      }
+      if (result != null) {
+        result.consume()
+      }
+    } finally {
```
jexp, line 57: should we catch exceptions here? as they are probably not relevant.

src/main/scala/org/neo4j/spark/utils/Neo4jUtils.scala
```diff
@@ -0,0 +1,58 @@
...
+        driver.close()
+      }
+    }
+  }
+
+  private val retryConfig = RetryConfig.custom.retryExceptions(
+      classOf[SessionExpiredException], classOf[ServiceUnavailableException] // retry on the same exceptions the driver does [1]
+    )
+    .retryOnException(new function.Predicate[Throwable] {
+      override def test(exception: Throwable): Boolean = if (exception.isInstanceOf[TransientException]) {
```
jexp, line 29: should probably a pattern match in the future

src/main/scala/org/neo4j/spark/utils/Neo4jUtils.scala
```diff
@@ -0,0 +1,58 @@
...
+        session.close()
+      }
+    } finally {
+      if (driver != null) {
+        driver.close()
+      }
+    }
+  }
+
+  private val retryConfig = RetryConfig.custom.retryExceptions(
```
jexp, line 25: wouldn't that be usually be something that's handled by Spark? How do others handle this? Just curious.

src/main/scala/org/neo4j/spark/dataframe/Neo4jDataFrame.scala
conker84: Just because I want to left the old implementation as much as possible in order to reduce the impact of the new changes.
If you want I can add the same strategy here and in the other parts

src/main/scala/org/neo4j/spark/dataframe/Neo4jDataFrame.scala
conker84: Just because otherwise this will not work:

https://github.com/neo4j-contrib/neo4j-spark-connector/blob/70ae1f647b6f66eddee666e55d048faded17f5f0/src/main/scala/org/neo4j/spark/Executor.scala#L24

src/main/scala/org/neo4j/spark/rdd/Neo4jTupleRDD.scala
conker84: Done in this one

src/main/scala/org/neo4j/spark/utils/Neo4jSessionAwareIterator.scala
conker84: Added a logger

src/main/scala/org/neo4j/spark/utils/Neo4jSessionAwareIterator.scala
conker84: What do you mean?

src/main/scala/org/neo4j/spark/utils/Neo4jUtils.scala
conker84: Added in this one

src/main/scala/org/neo4j/spark/utils/Neo4jUtils.scala
conker84: The spark retry mechanism works with `spark.task.maxFailures` which is:
```
Number of failures of any particular task before giving up on the job.
The total number of failures spread across different tasks will not cause the job to fail;
a particular task has to fail this number of attempts
```

But IMHO the control it's not fine-graded as this one, I would like to have both of them, wdyt?

conker84:
@jexp I updated the PR please check my answers to the comments.
Thanks a lot for the feedback!!!

src/main/scala/org/neo4j/spark/utils/Neo4jUtils.scala
```diff
@@ -0,0 +1,71 @@
...
+import org.neo4j.driver.exceptions.{ServiceUnavailableException, SessionExpiredException, TransientException}
+import org.neo4j.spark.Neo4jConfig
+
+
+object Neo4jUtils {
+
+  def close(driver: Driver, session: Session): Unit = {
+    try {
+      if (session != null && session.isOpen) {
+        session.close()
```
jexp, line 20: Please add the exception handling as we discussed.

src/test/scala/org/neo4j/spark/Neo4jSparkTSE.scala
```diff
@@ -85,18 +74,18 @@ class Neo4jSparkTest {
   }
 
   @Test def runCypherQueryWithPartition() {
-    val neo4j: Neo4j = Neo4j(sc).cypher("MATCH (n:Person) RETURN id(n) SKIP {_skip} LIMIT {_limit}").partitions(4).batch(25)
+    val neo4j: Neo4j = Neo4j(sc).cypher("MATCH (n:Person) RETURN id(n) SKIP $_skip LIMIT $_limit").partitions(4).batch(25)
```
jexp, line 77: did we also change the test inside of the execution/partition?
and the docs in the readme?

src/test/scala/org/neo4j/spark/SparkConnectorScalaBaseTSE.scala
```diff
@@ -0,0 +1,64 @@
...
+  val _testName: TestName = new TestName
+
+  @Rule
+  def testName = _testName
+
+  @Before
+  def before() {
+    SparkConnectorScalaSuiteIT.session()
+      .writeTransaction(new TransactionWork[ResultSummary] {
+        override def execute(tx: Transaction): ResultSummary = tx.run("MATCH (n) DETACH DELETE n").consume()
```
jexp, line 49: does this work with larger test datasets?

conker84: no, we must use a different strategy

src/test/scala/org/neo4j/spark/Neo4jSparkTSE.scala
conker84: yes in both of them

src/main/scala/org/neo4j/spark/utils/Neo4jUtils.scala
conker84: done

README.md
```diff
@@ -14,29 +14,29 @@ This neo4j-spark-connector is Apache 2 Licensed
 ## Building
 
 
-Build `target/neo4j-spark-connector_2.11-full-2.4.1-M1.jar` for Scala 2.11
+Build `target/neo4j-spark-connector_2.11-full-2.6.0-M1.jar` for Scala 2.11
```
jexp, line 17: 2.4.0-M2

jexp: 2.4.4-M2

pom.xml
```diff
@@ -4,7 +4,7 @@
     <modelVersion>4.0.0</modelVersion>
     <groupId>neo4j-contrib</groupId>
     <packaging>jar</packaging>
-    <version>2.4.1-M1</version>
+    <version>2.6.0-M1</version>
```
jexp, line 7: 2.4.4-M2

src/main/scala/org/neo4j/spark/dataframe/Neo4jDataFrame.scala
```diff
@@ -287,12 +212,11 @@ object CypherTypes {
...
     else if (typ == typeSystem.STRING()) CypherTypes.STRING
     else if (typ == typeSystem.INTEGER()) CypherTypes.INTEGER
-    else if (typ == typeSystem.FLOAT()) CypherTypes.FlOAT
     else if (typ == typeSystem.DATE()) CypherTypes.DATE
     else if (typ == typeSystem.DATE_TIME()) CypherTypes.DATETIME
     else if (typ == typeSystem.LOCAL_DATE_TIME()) CypherTypes.DATETIME
-    else if (typ == typeSystem.TIME()) CypherTypes.TIME
-    else if (typ == typeSystem.LOCAL_TIME()) CypherTypes.TIME	
+    else if (typ == typeSystem.TIME()) CypherTypes.DATETIME
+    else if (typ == typeSystem.FLOAT()) CypherTypes.FlOAT
```
jexp, line 219: where did typeSystem.LOCAL_TIME go?

src/main/scala/org/neo4j/spark/utils/Neo4jSessionAwareIterator.scala
```diff
@@ -0,0 +1,70 @@
...
+  private def close(rollback: Boolean = false) = {
+    try {
+      if (transaction != null && transaction.isOpen) {
+        if (rollback && write) {
+          transaction.rollback()
+        } else {
+          transaction.commit()
+        }
+      }
+      if (result != null) {
```
jexp, line 60: are you sure that this should happen after tx.commit / rollback? I'm not sure, perhaps ask Nigel?

src/main/scala/org/neo4j/spark/utils/Neo4jUtils.scala
jexp: perhaps make it configurable? (later PR)

src/test/scala/org/neo4j/spark/Neo4jDataFrameScalaTSE.scala
```diff
@@ -44,9 +28,8 @@ class Neo4jDataFrameScalaTest {
     val df = new SQLContext(sc).createDataFrame(rows, schema)
     Neo4jDataFrame.mergeEdgeList(sc, df, ("Person",Seq("name")),("ACTED_IN",Seq.empty),("Movie",Seq("title")))
 
-    val it: ResourceIterator[Long] = server.graph().execute("MATCH (:Person {name:'Keanu'})-[:ACTED_IN]->(:Movie {title:'Matrix'}) RETURN count(*) as c").columnAs("c")
-    assertEquals(1L, it.next())
-    it.close()
+    val count = SparkConnectorScalaSuiteIT.session().run("MATCH (:Person {name:'Keanu'})-[:ACTED_IN]->(:Movie {title:'Matrix'}) RETURN count(*) as c").single().get("c").asLong()
```
jexp, line 31: consider using triple-quotes for these long statements, but that can also happen in a later PR

PR #128, opened 2020-05-11
Documentation Structure

moxious:
- Basic documentation structure that matches the approach of neo4j-streams
- Very little content right now, but will add to it time allowing

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,4 @@ out
 Thumbs.db
 .cache-main
 .cache-tests
+docs/guides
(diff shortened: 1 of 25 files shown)


conker84:
there are several references to the streams doc I think that we can merge it as-is and improve in a different PR. @moxious wdyt?

moxious:
that was the intent -- just to merge this as a "structure" PR and not worry about the content . The content doesn't much matter until we set up the build chain to publish the resulting HTML out to a site.  This can come with an improvement PR.  I've been trying to get enough free time to collect all of the blog posts that are out there and at least write up quick start documentation and so on.  I just didn't want the docs structure to sit too long if I was running into problems finding the time to write real content, that's why it's a bit empty

conker84:
perfect, LGTM

PR #134, opened 2020-06-29
Updated saveGraph.updateNodes to use numPartitions instead of batch size.

nielsdejong:
Bug discovered during customer PoC, seems we are using **batchSize** where **numPartitions** should be.

### DIFF
--- a/src/main/scala/org/neo4j/spark/Neo4jGraph.scala
+++ b/src/main/scala/org/neo4j/spark/Neo4jGraph.scala
@@ -104,8 +104,8 @@ object Neo4jGraph {
           }) +
             s" $matchMerge (n)-[rel:${relType.quote}]->(m) SET rel.${relProp.quote} = row.value return count(*)"
         val batchSize = ((graph.numEdges / 100) + 1).toInt
-
-        graph.edges.repartition(batchSize).mapPartitions[Long](
+        val numPartitions = ((graph.numEdges / batchSize) + 1).toInt
+        graph.edges.repartition(numPartitions).mapPartitions[Long](
           p => {
             val rows = p.map(e => Seq(("from", e.srcId), ("to", e.dstId), ("value", e.attr)).toMap.asJava).toList.asJava
             val res1 = execute(config, updateRels, Map("data" -> rows)).rows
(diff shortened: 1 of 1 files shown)


conker84:
LGTM thanks a lot!

jexp:
@conker84 did you test it locally so that no test breaks? Also should probably add a test that validates this.

conker84:
> @conker84 did you test it locally so that no test breaks? Also should probably add a test that validates this.

@jexp Yes I tested locally I need to investigate why the tests are failing on Travis, I was wondering if makes sense add this kind of change `val numPartitions = ((graph.numEdges / batchSize) + 1).toInt` also after line 105:

https://github.com/neo4j-contrib/neo4j-spark-connector/blob/3991dee7d182ab17dd4f7057ee0672147b6a68ff/src/main/scala/org/neo4j/spark/Neo4jGraph.scala#L91-L117

@nielsdejong can you please add a test case?

jexp:
@conker84 yes I think so, that might be a 2nd place where to fix this.

conker84:
@nielsdejong can you please update the change also for the relationships?

nielsdejong:
@conker84 @jexp sure, will fix the change also for relationships & add tests. I'll have some time next week to work this out :)

nielsdejong:
hi @conker84 , @jexp - I've update the other line to also use the right number of partitions.
I'm currently working on a test `saveGraphValidatePartitions` in `Neo4jGraphScalaTSE` that checks whether the right number of partitions are created.

My current thought is to check the Neo4j logs for the number of write transactions created. What would be the best way to approach this?

 ```
 @Test def saveGraphValidatePartitions {
    val edges : RDD[Edge[VertexId]] = sc.makeRDD(...) // create graph with edges nodes
    val graph = Graph.fromEdges(edges,-1)
    Neo4jGraph.saveGraph(sc,graph,null,("REL","test"))
    // assert that the right number of partitions were created.  
  }
```

conker84:
@nielsdejong good point, I guess that given the complexity on testing this we can simply omit the test this time.
Thank a lot!
@jexp LGTM

conker84:
I'm merging this thank you @nielsdejong !!!

PR #147, opened 2020-07-10
Fixes #139: Create Option Class in order to connect to Neo4j

### LINKED ISSUE #139: Create Option Class in order to connect to Neo4j
We want to create a shared class between read/write-support that allows to connect to Neo4j in every scenario like we did in the Neo4j Streams Kafka Connect Plugin:
https://github.com/neo4j-contrib/neo4j-streams/blob/3.5/kafka-connect-neo4j/src/main/kotlin/streams/kafka/connect/sink/Neo4jSinkConnectorConfig.kt

utnaf:
Fixes #139 

Added Option class to handle query configurations and driver parameters. 

Features:
- check for required optins
- check for incompatible options

**Note**: Options list is not exhaustive, more parameters will be added when needed.

### DIFF
--- a/src/main/scala/org/neo4j/spark/Neo4jOptions.scala
+++ b/src/main/scala/org/neo4j/spark/Neo4jOptions.scala
@@ -24,14 +24,14 @@ class Neo4jOptions(private val parameters: java.util.Map[String, String]) {
     parameters.get(parameter).trim()
   }
 
-  val query: QueryOption = (
+  val query: Neo4jQueryOptions = (
     getParameter(QUERY.toString.toLowerCase),
     getParameter(NODE.toString.toLowerCase),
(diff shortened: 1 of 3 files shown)


conker84:
LGTM
@moxious @jexp if is fine with you I can merge it

jexp:
Is this compatible with the existing code?
If not we need a migration approach and depreciation of the previous approach 

Is this in line with other spark extensions?

And would it work with the future python/R apis?

conker84:
> Is this compatible with the existing code?
> If not we need a migration approach and depreciation of the previous approach

No, this is for the new APIs

> Is this in line with other spark extensions?
> 
> And would it work with the future python/R apis?

Yes this a DTO used to exchange data between spark and the Neo4j Driver

PR #148, opened 2020-07-17
Migrated documentation to Antora

mroiter-larus:
In order to locally preview the Neo4j Spark connector documentation built with Antora do the following steps:

1. open a terminal window and be sure to be at the root of the project
2. run the following command: `cd doc`
3. run the following command: `antora docs.yml`
4. run the following command: `npm start`
5. open a browser window and go to `localhost:8000`

### DIFF
--- a/doc/docs/antora.yml
+++ b/doc/docs/antora.yml
@@ -1,0 +1,12 @@
+name: neo4j-spark-docs
+version: 1.0.0
+title: Neo4j Spark Docs
+start_page: ROOT:index.adoc
+nav:
+  - modules/ROOT/nav.adoc
+
+asciidoc:
+  attributes:
+    theme: docs
+    docs-version: 1.0.0
+    copyright: Neo4j Inc.
--- a/doc/docs/modules/ROOT/pages/index.adoc
+++ b/doc/docs/modules/ROOT/pages/index.adoc
@@ -1,0 +1,19 @@
+
+= Neo4j Spark Connector v{docs-version}
+
+ifdef::backend-html5[(C) {copyright}]
+
+License: link:{attachmentsdir}/LICENSE.txt[Creative Commons 4.0]
+
+[abstract]
+--
+This is the user guide for Neo4j Spark Connector version {docs-version}, authored by the Neo4j Labs Team.
+--
+
+The guide covers the following areas:
+
+* xref:overview.adoc[Project overview]
+* xref:quickstart.adoc[Get Started Fast with the most Common Scenarios]
+* xref:neo4j-cluster.adoc[Using with Neo4j Causal Cluster]
+* xref:cloud.adoc[Configuring a connection to a Confluent Cloud]
+* xref:faq.adoc[Neo4j Spark FAQ]
--- a/doc/docs/modules/ROOT/pages/neo4j-cluster.adoc
+++ b/doc/docs/modules/ROOT/pages/neo4j-cluster.adoc
@@ -1,0 +1,28 @@
+
+[#neo4j_causal_cluster]
+= Using with Neo4j Causal Cluster
+
+[abstract]
+--
+This chapter describes considerations around using Neo4j Spark Connector with Neo4j Enterprise Causal Cluster.
+--
+
+== Overview
+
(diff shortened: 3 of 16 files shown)


moxious:
Tiny note, in the instructions, of course you have to do npm install in the docs folder before `npm start` will work.

But the actual online docs generated by this look *great*.

I'm OK to merge this.  Sounds like you guys are in contact with Adam about the structure, and as far as editing & writing new docs, it looks just as easy as the old method, so from a writing perspective it doesn't even seem like a big change (just layout / production)

mroiter-larus:
@moxious you're right! I forgot to mention that also a `npm install` (and in particular `npm install express`) is needed.

doc/docs/antora.yml
```diff
@@ -0,0 +1,11 @@
+name: docs
```
Mogztter, line 1: We should use a unique and descriptive name because Antora is using the component name in resource IDs and when generating the URLs.
For instance, if we want to do a cross-reference:

```
See xref:1.0.0@spark-connector:ROOT:overview.adoc[].
```

doc/docs/modules/ROOT/pages/index.adoc
```diff
@@ -0,0 +1,37 @@
...
+The guide covers the following areas:
+
+* <<_project_overview>> -- Project overview
+* <<_quick_start>> -- Get Started Fast with the most Common Scenarios
+* <<neo4j_causal_cluster>> -- Using with Neo4j Causal Cluster
+* <<cloud>> -- Configuring a connection to a Confluent Cloud
+* <<faq>> -- Neo4j Streams FAQ
+
+
+include::overview.adoc[]
```
Mogztter, line 29: I don't think we need to include each pages in the index. It will be possible to navigate between each pages using the left navigation.

doc/docs/modules/ROOT/pages/neo4j-cluster.adoc
```diff
@@ -0,0 +1,30 @@
+
+[#neo4j_causal_cluster]
+== Using with Neo4j Causal Cluster
```
Mogztter, line 3: If we don't include the pages then we should start the pages with a document title (i.e. in this case `= Using with Neo4j Causal Cluster`)

doc/docs/modules/ROOT/pages/index.adoc
```diff
@@ -0,0 +1,37 @@
+
+= Neo4j Spark Connector v{docs-version}
+:toc:
+:toclevels: 5
+:sectids:
+:sectlinks:
+:sectnums:
+:sectnumlevels: 5
+:env-docs: true
```
Mogztter, line 9: I think it should be defined in the playbook otherwise the conditional blocks are not very useful because the value will always be true

Mogztter:
Please note that Antora was designed to work using the `file://` protocol so technically you don't need a local HTTP server to serve the documentation site but using one is totally fine 👌

conker84:
How can we integrate this into our release pipeline? It's something that we can delegate to maven?

Mogztter:
@conker84 I guess you could call the `antora` CLI from Maven but you will still need to install Node.js and run `npm i` to install the dependencies.

For Gradle, there are a few options but, as far as I know, there are not "mature" yet:
- https://github.com/rwinch/antora-gradle-plugin
- https://ysb33rorg.gitlab.io/nodejs-gradle-plugin/0.9.3/docs/product-documentation.html

mroiter-larus:
> Please note that Antora was designed to work using the `file://` protocol so technically you don't need a local HTTP server to serve the documentation site but using one is totally fine 👌

@Mogztter I know that Antora was designed to work using `file://` protocol, but i wasn't able to correctly preview the documentation without a local HTTP server, because it seemed Antora didn't recognize the CSS

conker84:
@moxious I'm merging this, further improvements will be provided in new PRs.
@mroiter-larus thank you!

conker84:
and @Mogztter thank you for the review!

Mogztter:
You're welcome, good work :+1:

PR #151, opened 2020-07-23
Fixes #141: Add support to read from nodes with provided labels

### LINKED ISSUE #141: [ReadSupport] Add support to read from nodes with provided labels
We want read nodes as tables in the following way:

```scala
val simpleDf = sparkSession
   .read
   .format("neo4j")
   .option("url", "neo4j://localhost:7687")
   .option("node", "Person")
   .load()
```

Example:
For the given set of nodes:
```cypher
CREATE (p:Person:Customer{name: 'Andrea', surname: 'Santurbano'})
CREATE (p:Person{name: 'Davide', surname: 'Fantuzzi'})
```

The output of the Dataframe will be the following:

`<id>` | `<labels>` | `name` | `surname`
-- | -- | -- | --
0 | [Person,Customer] | Andrea | Santurbano
1| [Person] | Davide | Fantuzzi

N.b. this feature will support push-down filtering in order to optimize the data moved from Neo4j to Spark

utnaf:
Fixes #141 

This is a big one, we needed to write all the support objects, but these will also serve for the other issues so the next PRs will be hopefully smaller.

The `Getting Started` doc page now contains a draft of the docs for this version of the connector to better understand what's going. We'll refine the docs once we have most of the code done.


src/main/scala/org/neo4j/spark/Neo4jQuery.scala
```diff
@@ -0,0 +1,27 @@
...
+  val NODE_ALIAS = "n"
+  val INTERNAL_ID_FIELD = "<id>"
+  val INTERNAL_LABELS_FIELD = "<labels>"
+
+  private val renderer: Renderer = Renderer.getDefaultRenderer
+
+  def build(queryOptions: Neo4jQueryOptions): String =
+    queryOptions.queryType match {
+      case LABELS =>
+        val labels: List[String] = queryOptions.value.split(":").toList
```
moxious, line 20: Labels are specified like this:   `Foo:Bar`.   That's good, looks very "neo4j-y".   When doing this and interpolating into queries though, remember that labels can have weird characters, and may be enclosed in backticks.  By making this a string, it encourages users to treat it like cypher.  

But what if I set labels to be  "\`Foo Bar\`:Baz" will this work?   As a user, I'd expect it to, since that's how I did it if I have a MATCH query.

src/main/scala/org/neo4j/spark/service/SchemaService.scala
```diff
@@ -0,0 +1,140 @@
...
+
+  private val session: Session = driverCache.getOrCreate().session(options.session.toNeo4jSession)
+
+  private def cypherToSparkType(cypherType: String): DataType = {
+    cypherType match {
+      case "Boolean" => DataTypes.BooleanType
+      case "String" => DataTypes.StringType
+      case "Long" => DataTypes.IntegerType
+      case "Double" => DataTypes.DoubleType
+      case "Point" | "InternalPoint2D" | "InternalPoint3D" => pointType
```
moxious, line 34: let's talk about this more on friday.  Type mapping is a critical piece.  No change needed -- but want to understand options here.  In particular neo4j points are a bit strange (having x, y, z, and scheme) and I'm not sure about the spark side of this.

src/main/scala/org/neo4j/spark/service/SchemaService.scala
```diff
@@ -0,0 +1,140 @@
...
+      .sortBy(t => t.name)
+
+    structFields += StructField(Neo4jQuery.INTERNAL_LABELS_FIELD, DataTypes.createArrayType(DataTypes.StringType), nullable = true)
+    structFields += StructField(Neo4jQuery.INTERNAL_ID_FIELD, DataTypes.IntegerType, nullable = false)
+    StructType(structFields.reverse)
+  }
+
+  private def retrieveSchemaFromApoc: mutable.Buffer[StructField] = {
+    session.run(
+      "CALL apoc.meta.nodeTypeProperties({ includeLabels: $labels })",
```
moxious, line 69: PLACEHOLDER -- doesn't have to be handled in this PR.   This particular call along with relTypeProperties() has other parameters that it takes which are important.  They control sampling and metadata exclusion, among other things.    I think we may want a setting that allows a user to pass parameters here, to control how the metadata sampling is done.  On very large stores, this can get intensive without the right settings.

utnaf: Totally agree! We are aware of this, we mentioned it in the [estimate doc](https://docs.google.com/document/d/1qG0Wf8whkns3u6UAgrQTVaxGerEyCZHThNdmkVr3-UA/edit#). There is a lot going on in this PR so had to keep something out (like partitioning).

src/main/scala/org/neo4j/spark/Neo4jOptions.scala
```diff
@@ -24,78 +29,165 @@ class Neo4jOptions(private val parameters: java.util.Map[String, String]) {
...
   val ENCRYPTION_CA_CERTIFICATE_PATH = "encryption.ca.certificate.path"
-  val MAX_CONNECTION_LIFETIME_MSECS = "connection.max.lifetime.msecs"
-  val MAX_CONNECTION_ACQUISITION_TIMEOUT_MSECS = "connection.acquisition.timeout.msecs"
+  val CONNECTION_MAX_LIFETIME_MSECS = "connection.max.lifetime.msecs"
+  val CONNECTION_LIVENESS_CHECK_TIMEOUT = "connection.liveness.timeout.msec"
+  val CONNECTION_ACQUISITION_TIMEOUT_MSECS = "connection.acquisition.timeout.msecs"
+  val CONNECTION_TIMEOUT_MSECS = "connection.timeout.msecs"
 
   // session options
   val DATABASE = "database"
```
moxious, line 175: I tested it, and get a NullPointerException if Database is specified as None in python.  If it's not a valid value, ignore the setting instead and user gets default database.

doc/docs/modules/ROOT/pages/quickstart.adoc
```diff
@@ -8,28 +8,274 @@ Get started fast for common scenarios, using Neo4j Streams plugin or Kafka Conne
...
+```
 
 .Most Common Needed Configuration Settings
 |===
-|Setting Name |Description |Default Value
+|Setting Name |Description |Default Value |Required 
+
+4+|*Driver Options*
+
+|`url`
```
moxious, line 46: I'm not sure about this one.  Normally in driver URLs you can specify the database, neo4j://localhost/foo

We may want to do it this way and entirely drop the database option, WDYT?   If we *don't* drop the database option, then we have to figure out what to do:

```
.option("url", "neo4j://localhost/foo")
.option("database", "bar")
```

doc/docs/modules/ROOT/pages/quickstart.adoc
```diff
@@ -8,28 +8,274 @@ Get started fast for common scenarios, using Neo4j Streams plugin or Kafka Conne
...
+
+|`connection.timeout.msecs`
+|Connection timeout in milliseconds
+|(Neo4j Driver default)
+|No
+
+
+4+|*Session Options*
+
+|`database`
```
moxious, line 129: see note on URL

doc/docs/modules/ROOT/pages/quickstart.adoc
```diff
@@ -8,28 +8,274 @@ Get started fast for common scenarios, using Neo4j Streams plugin or Kafka Conne
...
+|No
+
+4+|*Data Read Options*
+
+|`query`
+|Specify a query to read the data
+|``
+|Yes^*^
+
+|`labels`
```
moxious, line 146: I did this: 

```
  .option("labels", ":City") \
```

Maybe bit a bit more tolerant because this is a very natural thing for a Neo4j user to do, although technically we're passing a blank label, I know.

```
org.apache.spark.SparkException: Job aborted due to stage failure: Task 0 in stage 12.0 failed 1 times, most recent failure: Lost task 0.0 in stage 12.0 (TID 12, localhost, executor driver): java.lang.IllegalArgumentException: A primary label is required.
```

doc/docs/modules/ROOT/pages/quickstart.adoc
```diff
@@ -8,28 +8,274 @@ Get started fast for common scenarios, using Neo4j Streams plugin or Kafka Conne
...
+val df = spark.read.format("org.neo4j.spark.DataSource")
+        .option("url", "bolt://localhost:7687")
+        .option("labels", "Person:Customer:Admin")
+        .load()
+
+df.show()
+```
+
+[NOTE]
+Remember that the first Label is considered to be the primary label.
```
moxious, line 246: I don't know what "Primary label" means, or what impact it has.

src/main/scala/org/neo4j/spark/Neo4jOptions.scala
conker84: can you share the code and the stacktrace?

doc/docs/modules/ROOT/pages/quickstart.adoc
conker84: I wasn't aware of this, in fact here we used here the same approach as streams. I tried to look into the Java Driver docs but I didn't find anyting.
If you agree I would like to keep the same approach as the Kafka Connect Sink, so the behaviour in the case that you highlighted:
```
.option("url", "neo4j://localhost/foo")
.option("database", "bar")
```

Will use the bar as database `bar`. Otherwise will use `foo` so the explicit option has the "priority" respect to the URL. We can document it.
WDYT?

doc/docs/modules/ROOT/pages/quickstart.adoc
conker84: It' just how Cypher DSL handles the labels, we can remove this.

doc/docs/modules/ROOT/pages/quickstart.adoc
conker84: done

conker84:
thank you everybody!

PR #155, opened 2020-07-28
fixes #144: [WriteSupport] Write a Dataframe to Neo4j as nodes

### LINKED ISSUE #144: [WriteSupport] Write a Dataframe to Neo4j as nodes
We want to save a Dataframe as Nodes
```scala
otherFormatDf
   .write
   .format("neo4j")
   .mode(SaveMode.Overwrite)
   .option("node", "Person")
   .option("nodeKeys", "ssn") // comma separated list of keys
   .save()
```

For the given Dataframe:

ssn | name | surname
-- | -- | --
AB12 | Andrea | Santurbano
CD34 | Davide | Fantuzzi

Under-the-hood we'll perform the following query:

```cypher
UNWIND batch AS row
MERGE (p:Person{ssn: row.ssn}) SET p += row
```

We'll support two `SaveMode` (https://spark.apache.org/docs/2.4.4/api/java/index.html?org/apache/spark/sql/SaveMode.html):
* `Overwrite`: this will perform a `MERGE` query
* `ErrorIfExists`: this will perform a `CREATE` query

conker84:
This fixes #144 by adding the support for write nodes:

```scala
ds.write
  .mode(SaveMode.ErrorIfExists)
  .option("url", "bolt://localhost:7687")
  .option("labels", ":Person:Customer")
  .save()
```

or

```scala
ds.write
  .mode(SaveMode.Overwrite)
  .option("url", "bolt://localhost:7687")
  .option("labels", ":Person:Customer")
  .option("node.keys", "name,surname")
  .save()
```

As it depends on the #141 I put it draft until the related PR is merged


src/main/scala/org/neo4j/spark/Neo4jOptions.scala
```diff
@@ -188,6 +231,8 @@ object Neo4jOptions {
   val DEFAULT_ENCRYPTION_ENABLED = false
   val DEFAULT_ENCRYPTION_TRUST_STRATEGY = TrustStrategy.Strategy.TRUST_SYSTEM_CA_SIGNED_CERTIFICATES
   val DEFAULT_SCHEMA_FLATTEN_LIMIT = 10
+  val DEFAULT_BATCH_SIZE = 500
```
moxious, line 234: depends on node size of course but as a default I'd go much higher, like 5k.   Typical values we use as a starting point in some tests are 20k, but these can be dangerously high and blow up heap if nodes are fat.

src/main/scala/org/neo4j/spark/Neo4jOptions.scala
```diff
@@ -110,7 +144,7 @@ case class Neo4jDriverOptions(
                              ) extends Serializable {
 
   def toDriverConfig: Config = {
-    val builder = Config.builder()
+    val builder = Config.builder().withUserAgent("neo4j-spark-connector")
```
moxious, line 147: include version please.  Inevitably we'll have a bug later and will need to know what a customer is using

src/main/scala/org/neo4j/spark/writer/Neo4jDataWriter.scala
```diff
@@ -0,0 +1,115 @@
...
+  private val batch: util.List[java.util.Map[String, Object]] = new util.ArrayList[util.Map[String, Object]]()
+
+  private val retries = new CountDownLatch(options.transactionMetadata.retries)
+
+  val query: String = new Neo4jQueryService(options, new Neo4jQueryWriteStrategy(saveMode)).createQuery()
+
+  override def write(record: InternalRow): Unit = {
+    batch.add(mappingService.toParameter(record, structType))
+    if (batch.size() == options.nodeMetadata.batchSize) {
+      writeBatch
```
moxious, line 42: docs need reference: since fundamentally we're doing batched writes, if a late batch fails, this isn't transactional and all the previous batches still stay

src/main/scala/org/neo4j/spark/writer/Neo4jDataWriter.scala
```diff
@@ -0,0 +1,115 @@
...
+    }
+  }
+
+  private def writeBatch(): Unit = {
+    if (session == null || !session.isOpen) {
+      session = driverCache.getOrCreate().session(options.session.toNeo4jSession)
+    }
+    if (transaction == null || !transaction.isOpen) {
+      transaction = session.beginTransaction()
```
moxious, line 51: beginWriteTransaction?    Explicit

src/main/scala/org/neo4j/spark/writer/Neo4jDataWriter.scala
conker84: we can improve this in the future by adding a property like `batch.commit=3` that says commit every three batches (= 5k*3=15k rows in the default case). WDYT?

conker84: and we can also provide an `ingestion.mode=BATCH/TRANSACTIONAL` that can allow the use to choose the ingestion mode. WDYT?

src/main/scala/org/neo4j/spark/writer/Neo4jDataWriter.scala
conker84: we explicit it in the session with `AccesMode.WRITE`

src/main/scala/org/neo4j/spark/Neo4jOptions.scala
conker84: done

src/main/scala/org/neo4j/spark/writer/Neo4jDataWriter.scala
moxious: Fine point, but if you put 3 batches in a transaction, I think it still requires the memory of 3 batches.  So I don't see the point in putting 3 batches in 1 TX, why not just make 1 bigger batch.  The limiting factor is memory and the potential for the server to OOM.

As for transactional, it seems to me you don't need this mode, because it's the same as making the batch size more than the total number of records in the write.   I.e. make the entire write one batch.

I think we need to document how it works, but basic batching with an adjustable size is the right way to go.  It's simple, easy to implement, and also easy to understand.  If users don't like batching and want it to be transactional, they can just make the batch number really really large assuming they have the memory.

PR #156, opened 2020-07-30
 fixes #143: [WriteSupport] Write a dataframe to Neo4j via provided query

### LINKED ISSUE #143: [WriteSupport] Write a dataframe to Neo4j via provided query
We want to ingest the a dataframe into Neo4j via provided query in the following way:
```scala
otherFormatDf
   .write
   .format("neo4j")
   .option("query", "MERGE (p:Person {ssn: row.ssn}) SET p.name = row.fullname")
   .save()
```
where `row` is the Dataframe row value provided as map.

Under-the-hood we'll use the same approach as the Neo4j Streams plugin so we'll append the provided query to an UNWIND operation:

```cypher
UNWIND batch AS row
MERGE (p:Person {ssn: row.ssn}) SET p.name = row.fullname
```

The default behaviour will use the Dataframe partition size as batch size, otherwise the user can provide an option param where sets the batch size.

conker84:
This PR fixes #143 allowing the user to define a simple query in order to ingest a Spark's Dataset via Cypher query in the following way:

```scala
ds.write
  .option("url", "bolt://localhost:7687")
  .option("query", "CREATE (n:Person{fullName: event.name + event.surname})")
  .save()
```

### DIFF
--- a/doc/docs/modules/ROOT/pages/quickstart_writer.adoc
+++ b/doc/docs/modules/ROOT/pages/quickstart_writer.adoc
@@ -190,5 +190,3 @@ Under the hod the Spark Connector will flatten the data in this way:
 |NY
 
 |===
-
-
--- a/doc/docs/modules/ROOT/pages/quickstart_writer.adoc
+++ b/doc/docs/modules/ROOT/pages/quickstart_writer.adoc
@@ -190,3 +190,29 @@ Under the hod the Spark Connector will flatten the data in this way:
 |NY
 
 |===
+
+===== Query
+
+In case you use the option `query` the Spark Connector will persist the entire Dataset by using the provided query.
+The nodes will be sent to Neo4j in a batch of rows defined in the `batch.size` property and we will
+perform the under the hood un `UNWIND` operation over the batch.
+
+So given the following simple Spark program:
+
+----
+ds.write
+  .format("org.neo4j.spark.DataSource")
+  .option("url", "bolt://localhost:7687")
+  .option("query", "CREATE (n:Person{fullName: event.name + event.surname})")
+  .save()
+----
+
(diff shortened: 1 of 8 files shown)


moxious:
Testing & trying I did this:

```
%py

cities.write.format("org.neo4j.spark.DataSource") \
   .mode("errorIfExists") \
   .option("url", dbutils.widgets.get("url")) \
   .option("authentication.basic.username", dbutils.widgets.get("user")) \
   .option("authentication.basic.password", dbutils.widgets.get("password")) \
   .option("query", "CREATE (c:City:Spark) SET c += event") \
   .save()
```

I think "errorIfExists" and certain other modes may not be compatible with query.   This looks like it does a certain thing, but the code can't really guarantee I'm not creating duplicates, right?   I wonder about "errorIfExists" should we make this Neo4j's problem?  I.e. if you want that behavior, why not expect the user to create constraints that will fail create when something exists?

moxious:
Placeholder:  I need a way to run cypher from spark to do things like guarantee the existence of indexes & constraints, that isn't starting from the dataframe.

Not sure yet how this will work.  I can't do spark.write (that doesn't exist) but spark.read is wrong because creating an index is a write operation.

Screen Shot 2020-07-31 at 7 22 38 AM

moxious:
Wondering (for later) how we manage situations where we want to save dataframes but want the information back?  (Like count the number of relationships created)

Screen Shot 2020-07-31 at 7 46 10 AM

moxious:
another floating thought -- each time we open a java driver connection to the database, make sure to control the max number of connections in the pool.  The driver by default wants to open a whole pool, but it seems like in most usage scenarios the spark adapter can never make use of more than 1.  We can probably speed up execution times a slight bit and also avoid confusions or misuse of connections if we only have 1 open

PR #158, opened 2020-08-04
Fixed Antora documentation generation

mroiter-larus:
Fixed the following 2 problems with the Antora configuration:

- Antora builds correctly but browsing to `localhost:8000` we get an error: `Cannot GET /docs/1.0.0`
- Antora builds correctly but `build/site` folder remains empty (no playbook were generated)


conker84:
Thank you @mroiter-larus

PR #160, opened 2020-08-05
fixes #140: [ReadSupport] Add support for reading from Neo4j by a provided query

### LINKED ISSUE #140: [ReadSupport] Add support for reading from Neo4j by a provided query
We want to read data from Neo4j by allowing the developer to provide a simple Cypher query like:
```scala
val simpleDf = sparkSession
   .read
   .format("neo4j")
   .option("url", "neo4j://localhost:7687")
   .option("query", "MATCH (p:Person) RETURN p")
   .load()
```

In this case we should “flatten” the result as we already did in the Neo4j JDBC driver:
https://neo4j-contrib.github.io/neo4j-jdbc/#_flattening
and
https://github.com/neo4j-contrib/neo4j-jdbc/blob/4.0/neo4j-jdbc-bolt/src/main/java/org/neo4j/jdbc/bolt/BoltNeo4jResultSet.java#L154
The flattening level should be provided via option param.

Schema inferred automagically:
* Using APOC  “apoc.meta.nodeTypeProperties” and “apoc.meta.relTypeProperties”, we’ll make sure to provide access to APOC arguments via Spark options.
* If APOC are not present, we’ll use a prepared query like “MATCH (n:Person) RETURN n ORDER BY rand() LIMIT <flattening_value_provided_by_option_param>”

*N.b.* this feature will **not** support push-down filtering

conker84:
fixes #140 

We added the `query` option so the user can define a Cypher query like this:

```scala
val df = spark.read.format("org.neo4j.spark.DataSource")
      .option("query", "MATCH (n:Person) WITH n LIMIT 2 RETURN collect(n) AS nodes")
      .option("schema.strategy", "string")
      .load()

df.show()
```

Moreover, we added the property `schema.strategy` with `string` and `sample` (default) values, wi the first one all the fields returned by the query will be coerced to String.

### DIFF
--- a/doc/docs/modules/ROOT/pages/quickstart.adoc
+++ b/doc/docs/modules/ROOT/pages/quickstart.adoc
@@ -128,8 +128,9 @@ in case you set this option will have the priority compared to the one defined i
 |No
 
 |`access.mode`
-|Access mode
-|`write`
+|Used only while you're pulling data from Neo4j. In case is `read` the connector, in a cluster environment
+will route the requests to the followers, otherwise to the leader.
+|`read`
 |No
 
 4+|*Data Read Options*
@@ -168,6 +169,12 @@ every single node property as column prefixed by `source` or `target`
 |`schema.flatten.limit`
 |Number of records to be used to create the Schema (only if APOC are not installed)
 |`10`
+|No
+
+|`schema.strategy`
+|Strategy used by the connector in order to compute the Schema definition for the Dataset. Values `string` and
+`sample`. If `string` coerces all the properties to String otherwise it will try to sample the Neo4j's dataset.
+|`sample`
 |No
 
 |===
@@ -329,12 +336,13 @@ Will create this schema
 
 === Read data by Relationship Type
 
-You can both specify a Cypher Path in this way:
+You can specify a Cypher Path in this way:
 ```scala
 val df = spark.read.format("org.neo4j.spark.DataSource")
       .option("relationship", "BOUGHT")
       .option("relationship.source.labels", "Person")
       .option("relationship.target.labels", "Product")
+      .load()
 
 df.show()
 ```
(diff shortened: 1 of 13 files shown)


doc/docs/modules/ROOT/pages/quickstart.adoc
```diff
@@ -366,3 +374,34 @@ otherwise if `false`:
...
+You can specify a Cypher query in this way:
+```scala
+val df = spark.read.format("org.neo4j.spark.DataSource")
+      .option("query", "MATCH (n:Person) WITH n LIMIT 2 RETURN collect(n) AS nodes")
+      .load()
+
+df.show()
+```
+
+The struct of the Dataset returned by the query is influenced by the query itself, in this particular context it could happen
```
moxious, line 389: put a note here -- recommend that individual property fields be returned, rather than returning node, relationship, and path types.  This best maps to spark's type system and yields best results.   So for the cypher query, better than:

```
MATCH (p:Person) RETURN p
```

is

```
MATCH (p:Person)
RETURN id(p) as id, p.name as name
```

Update the quick example to show this kind of usage pattern to encourage people off the bat to do the right thing.

doc/docs/modules/ROOT/pages/quickstart.adoc
```diff
@@ -366,3 +374,34 @@ otherwise if `false`:
...
+val df = spark.read.format("org.neo4j.spark.DataSource")
+      .option("query", "MATCH (n:Person) WITH n LIMIT 2 RETURN collect(n) AS nodes")
+      .option("schema.strategy", "string")
+      .load()
+
+df.show()
+```
+
+This means that the struct returned by the query will be composed by strings that you can than cast via simply Spark's
+transformations.
```
moxious, line 403: add note.   Inference is good when all instances of a property in neo4j are the same type, and string followed by cast is better when property types may differ.  Remember that Neo4j does not enforce property typing, and so person.age could sometimes be a long and sometimes be a string.

doc/docs/modules/ROOT/pages/quickstart.adoc
conker84: Done

doc/docs/modules/ROOT/pages/quickstart.adoc
conker84: Done

PR #162, opened 2020-08-10
Fixes issue #161: Add PushDownFilters support

### LINKED ISSUE #161: Add support for PushDownFilters
Add support for PushDownFilters as stated in #142 and #141 

We decided to do a different PR for these features to avoid having too many in the first one. 

Here's the doc for [SupportPushDownFilters](https://jaceklaskowski.gitbooks.io/mastering-spark-sql/content/spark-sql-SupportsPushDownFilters.html).

utnaf:
Fixes #161 

Added PushDownFilters support. 

Is applied just with `labels` and `relationship`, it apply with `relationship` and the `relationship.nodes.map` options set to `true`. 

When applied the filters are executed on the Cypher query, moving less data from Neo4j to Spark. When not applied the filters are executed by Spark.

It can be manually disabled setting `pushdown.filters.enabled` option to `false`.

### DIFF
--- a/doc/docs/modules/ROOT/pages/quickstart.adoc
+++ b/doc/docs/modules/ROOT/pages/quickstart.adoc
@@ -211,7 +211,7 @@ When using `relationship.node.map = true` or `query` the PushDownFilters support
 
 ==== How we extract the schema
 
-As Neo4j has ha schema-less approach and Spark needs a Schema in order to create a Dataset,
+As Neo4j has a schema-less approach and Spark needs a Schema in order to create a Dataset,
 we use several approaches in order to sample the dataset into Neo4j and compute the schema for Spark's Dataset.
 
 ===== Extract schema for Nodes
--- a/doc/docs/modules/ROOT/pages/quickstart.adoc
+++ b/doc/docs/modules/ROOT/pages/quickstart.adoc
@@ -408,12 +408,6 @@ df.show()
 This means that the struct returned by the query will be composed by strings that you can than cast via simply Spark's
 transformations.
 
-[NOTE]
-Inference (`schema.strategy` = `sample`) is good when all instances of a property in neo4j are the same type,
-and string followed by cast is better when property types may differ.
-Remember that Neo4j does not enforce property typing, and so `person.age` could sometimes be a `long
-and sometimes be a `string`.
-
 ==== Schema
 
 If APOC are installed, schema will be created with `apoc.meta.relTypeProperties`. Otherwise the first 10 (or any number specified by the `schema.flatten.limit` option) results will be flattened and the schema will be create from those properties.
--- a/doc/docs/modules/ROOT/pages/quickstart.adoc
+++ b/doc/docs/modules/ROOT/pages/quickstart.adoc
@@ -177,6 +177,11 @@ every single node property as column prefixed by `source` or `target`
 |`sample`
 |No
 
+|`pushdown.filters.enabled`
+|Enable or disable the Push Down Filters support
+|`true`
+|No
+
 |===
 
 ^*^ Just one of the options can be specified.
@@ -194,6 +199,15 @@ Reading data from a Neo4j Database can be done in 3 ways:
 Spark works with data in a tabular fixed schema. To accomplish this Neo4j Connector has a schema infer system that creates the schema based on the data requested for the read. Each read data method has is own strategy to create it, that will be explained it each section.
 
 TK list of supported data types
+
+=== Consideration on the filters
+
+The Neo4j Spark Connector implements the SupportPushDownFilters interface, that allows you to push the Spark filters down to the Neo4j layer. In this way the data that Spark will receive will be already filtered by Neo4j.
+
+You can manually disable the Push Down Filters support using the `pushdown.filters.enabled` option and set it to `false` (default is `true`).
+
+[NOTE]
+When using `relationship.node.map = true` or `query` the PushDownFilters support is not active, thus the filters will be applied by Spark and not by Neo4j.
 
 ==== How we extract the schema
 
@@ -388,8 +402,6 @@ If `relationship.node.map` is set to **true**
 * ``\`<target>`.\`[property]` `` for the target node map property
 
 in this case, all the map values will be strings, so the filter value must be a string too.
-
-**Note**: when using `relationship.node.map = true` the PushDownFilters support is not active, thus the filters will be applied by Spark and not by Neo4j.
 
 ```scala
 val df = spark.read.format("org.neo4j.spark.DataSource")
--- a/doc/docs/modules/ROOT/pages/quickstart.adoc
+++ b/doc/docs/modules/ROOT/pages/quickstart.adoc
@@ -371,9 +371,36 @@ otherwise if `false`:
 * `<targetId>` the internal Neo4j id of target node
 * `<targetLabels>` a list of labels for target node
 
+==== Filter
+
+You can use Spark to filter properties of the relationship, the source node, or the target node. Just use the correct prefix:
+
+If `relationship.node.map` is set to **false**
+
+* `source.[property]` for the source node properties
+* `rel.[property]` for the relation property
+* `target.[property]` for the target node property
+
+If `relationship.node.map` is set to **true**
+
+* ``\`<source>`.[property]`` for the source node map properties
+* ``\`<rel>`.[property]`` for the relation map property
+* ``\`<target>`.[property]`` for the target node map property
+
+```scala
+val df = spark.read.format("org.neo4j.spark.DataSource")
+      .option("relationship.node.map", false)
+      .option("relationship", "BOUGHT")
+      .option("relationship.source.labels", "Person")
+      .option("relationship.target.labels", "Product")
+      .load()
+
+df.where("source.name = 'John Doe' AND target.price >= 33")
+```
+
 ==== Schema
 
-If APOC are installed, schema will be created with `apoc.meta.relTypeProperties`. Otherwise the first 10 (or any number specified by the `schema.flatten.limit` option) results will be flattened and the schema will be create from those properties.
+If APOC are available, the schema will be created with `apoc.meta.relTypeProperties`. Otherwise the first 10 (or any number specified by the `schema.flatten.limit` option) results will be flattened and the schema will be create from those properties.
 
 === Read data by custom Cypher Query
 
--- a/doc/docs/modules/ROOT/pages/quickstart.adoc
+++ b/doc/docs/modules/ROOT/pages/quickstart.adoc
@@ -377,15 +377,19 @@ You can use Spark to filter properties of the relationship, the source node, or 
 
 If `relationship.node.map` is set to **false**
 
-* `source.[property]` for the source node properties
-* `rel.[property]` for the relation property
-* `target.[property]` for the target node property
+* ``\`source.[property]` `` for the source node properties
+* ``\`rel.[property]` `` for the relation property
+* ``\`target.[property]` `` for the target node property
 
 If `relationship.node.map` is set to **true**
 
-* ``\`<source>`.[property]`` for the source node map properties
-* ``\`<rel>`.[property]`` for the relation map property
-* ``\`<target>`.[property]`` for the target node map property
+* ``\`<source>`.\`[property]` `` for the source node map properties
+* ``\`<rel>`.\`[property]` `` for the relation map property
+* ``\`<target>`.\`[property]` `` for the target node map property
+
+in this case, all the map values will be strings, so the filter value must be a string too.
+
+**Note**: when using `relationship.node.map = true` the PushDownFilters support is not active, thus the filters will be applied by Spark and not by Neo4j.
 
 ```scala
 val df = spark.read.format("org.neo4j.spark.DataSource")
--- a/doc/docs/modules/ROOT/pages/quickstart.adoc
+++ b/doc/docs/modules/ROOT/pages/quickstart.adoc
@@ -439,6 +439,12 @@ df.show()
 This means that the struct returned by the query will be composed by strings that you can than cast via simply Spark's
 transformations.
 
+[NOTE]
+Inference (`schema.strategy` = `sample`) is good when all instances of a property in neo4j are the same type,
+and string followed by cast is better when property types may differ.
+Remember that Neo4j does not enforce property typing, and so `person.age` could sometimes be a `long
+and sometimes be a `string`.
+
 ==== Schema
 
 If APOC are installed, schema will be created with `apoc.meta.relTypeProperties`. Otherwise the first 10 (or any number specified by the `schema.flatten.limit` option) results will be flattened and the schema will be create from those properties.
--- a/doc/docs/modules/ROOT/pages/quickstart.adoc
+++ b/doc/docs/modules/ROOT/pages/quickstart.adoc
@@ -206,6 +206,12 @@ The Neo4j Spark Connector implements the SupportPushDownFilters interface, that 
 
 You can manually disable the Push Down Filters support using the `pushdown.filters.enabled` option and set it to `false` (default is `true`).
 
+If you use use the filter function more than once, like in this example:
+```scala
+df.where("name = 'John Doe'").where("age = 32")
+```
+The conditions will be automatically joined with an `AND` operator.
+
 [NOTE]
 When using `relationship.node.map = true` or `query` the PushDownFilters support is not active, thus the filters will be applied by Spark and not by Neo4j.
 
@@ -321,7 +327,8 @@ When reading data with this method, the Dataframe will contain all the fields co
 
 ==== Schema
 
-If APOC are installed, schema will be created with `apoc.meta.nodeTypeProperties`. Otherwise the first 10 (or any number specified by the `schema.flatten.limit` option) results will be flattened and the schema will be create from those properties.
+If APOC are available, the schema will be created with `apoc.meta.nodeTypeProperties`.
+Otherwise the first 10 (or any number specified by the `schema.flatten.limit` option) results will be flattened and the schema will be create from those properties.
 
 ===== Example
 
@@ -416,7 +423,8 @@ df.where("source.name = 'John Doe' AND target.price >= 33")
 
 ==== Schema
 
-If APOC are available, the schema will be created with `apoc.meta.relTypeProperties`. Otherwise the first 10 (or any number specified by the `schema.flatten.limit` option) results will be flattened and the schema will be create from those properties.
+If APOC are available, the schema will be created with `apoc.meta.relTypeProperties`. 
+Otherwise the first 10 (or any number specified by the `schema.flatten.limit` option) results will be flattened and the schema will be create from those properties.
 
 === Read data by custom Cypher Query
 
--- a/src/main/scala/org/neo4j/spark/service/Neo4jQueryService.scala
+++ b/src/main/scala/org/neo4j/spark/service/Neo4jQueryService.scala
@@ -38,7 +38,7 @@ class Neo4jQueryWriteStrategy(private val saveMode: SaveMode) extends Neo4jQuery
   }
 }
 
-class Neo4jQueryReadStrategy(filters: Array[Filter]) extends Neo4jQueryStrategy {
+class Neo4jQueryReadStrategy(filters: Array[Filter] = Array.empty[Filter]) extends Neo4jQueryStrategy {
   private val renderer: Renderer = Renderer.getDefaultRenderer
 
   override def createStatementForQuery(options: Neo4jOptions): String = options.query.value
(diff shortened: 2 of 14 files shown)


doc/docs/modules/ROOT/pages/quickstart.adoc
```diff
@@ -195,6 +200,15 @@ Spark works with data in a tabular fixed schema. To accomplish this Neo4j Connec
...
+The Neo4j Spark Connector implements the SupportPushDownFilters interface, that allows you to push the Spark filters down to the Neo4j layer. In this way the data that Spark will receive will be already filtered by Neo4j.
+
+You can manually disable the Push Down Filters support using the `pushdown.filters.enabled` option and set it to `false` (default is `true`).
+
+[NOTE]
+When using `relationship.node.map = true` or `query` the PushDownFilters support is not active, thus the filters will be applied by Spark and not by Neo4j.
+
 ==== How we extract the schema
 
 As Neo4j has ha schema-less approach and Spark needs a Schema in order to create a Dataset,
```
jexp, line 214: a

doc/docs/modules/ROOT/pages/quickstart.adoc
```diff
@@ -371,9 +385,38 @@ otherwise if `false`:
...
+      .option("relationship.target.labels", "Product")
+      .load()
+
+df.where("source.name = 'John Doe' AND target.price >= 33")
+```
+
 ==== Schema
 
-If APOC are installed, schema will be created with `apoc.meta.relTypeProperties`. Otherwise the first 10 (or any number specified by the `schema.flatten.limit` option) results will be flattened and the schema will be create from those properties.
+If APOC are available, the schema will be created with `apoc.meta.relTypeProperties`. Otherwise the first 10 (or any number specified by the `schema.flatten.limit` option) results will be flattened and the schema will be create from those properties.
```
jexp, line 419: One sentence per line please. Makes it easier for diffs and is the rule for Neo4j related docs (asciidoc)

src/main/scala/org/neo4j/spark/service/Neo4jQueryService.scala
```diff
@@ -37,24 +38,70 @@ class Neo4jQueryWriteStrategy(private val saveMode: SaveMode) extends Neo4jQuery
...
+
+    val relationship = sourceNode.relationshipBetween(targetNode, options.relationshipMetadata.relationshipType)
+      .named(Neo4jUtil.RELATIONSHIP_ALIAS)
+
+    val matchQuery = Cypher.`match`(sourceNode).`match`(targetNode).`match`(relationship)
+
+    if (filters.nonEmpty) {
+      val filtersMap: Map[PropertyContainer, Array[Filter]] = filters.map(filter => {
+        if (filter.isAttribute(Neo4jUtil.RELATIONSHIP_SOURCE_ALIAS)) {
+          (sourceNode, filter)
```
jexp, line 58: perhaps nicer to work with a case class

src/main/scala/org/neo4j/spark/service/Neo4jQueryService.scala
```diff
@@ -37,24 +38,70 @@ class Neo4jQueryWriteStrategy(private val saveMode: SaveMode) extends Neo4jQuery
...
+        }
+        else if (filter.isAttribute(Neo4jUtil.RELATIONSHIP_ALIAS)) {
+          (relationship, filter)
+        }
+        else {
+          throw new IllegalArgumentException(s"Attribute '${filter.getAttribute.get}' is not valid")
+        }
+      }).groupBy[PropertyContainer](_._1).mapValues(_.map(_._2))
+
+      matchQuery.where(
```
jexp, line 71: hard to reason about what happens in this code block

doc/docs/modules/ROOT/pages/quickstart.adoc
```diff
@@ -195,6 +200,15 @@ Spark works with data in a tabular fixed schema. To accomplish this Neo4j Connec
 
 TK list of supported data types
 
+=== Consideration on the filters
+
+The Neo4j Spark Connector implements the SupportPushDownFilters interface, that allows you to push the Spark filters down to the Neo4j layer. In this way the data that Spark will receive will be already filtered by Neo4j.
```
jexp, line 205: should probably mention that all filters are ANDed together?

src/main/scala/org/neo4j/spark/service/Neo4jQueryService.scala
```diff
@@ -37,24 +38,70 @@ class Neo4jQueryWriteStrategy(private val saveMode: SaveMode) extends Neo4jQuery
...
+        }.reduce { (a, b) => a.and(b) }
+      )
+    }
+
+    renderer.render(matchQuery.returning(node).build())
+  }
+
+  private def createNode(name: String, labels: Seq[String]) = {
+    val primaryLabel = labels.head
+    val otherLabels = labels.takeRight(labels.size - 1)
```
jexp, line 103: labels.tail()

src/main/scala/org/neo4j/spark/util/Neo4jImplicits.scala
```diff
@@ -56,4 +58,26 @@ object Neo4jImplicits {
...
+      case gt: GreaterThan => gt.attribute
+      case gte: GreaterThanOrEqual => gte.attribute
+      case lt: LessThan => lt.attribute
+      case lte: LessThanOrEqual => lte.attribute
+      case in: In => in.attribute
+      case notNull: IsNotNull => notNull.attribute
+      case isNull: IsNull => isNull.attribute
+      case startWith: StringStartsWith => startWith.attribute
+      case endsWith: StringEndsWith => endsWith.attribute
+      case contains: StringContains => contains.attribute
```
jexp, line 75: what about list data type (IN) and boolean values (no comparion where f.flag ) and temporal / spatial values (distance, components, ...)? perhaps in a later PR?

src/main/scala/org/neo4j/spark/util/Neo4jUtil.scala
```diff
@@ -205,15 +206,38 @@ object Neo4jUtil {
...
+    }
+  })
     .toMap
     .asJava
 
   def connectorVersion: String = properties.getOrDefault("version", "UNKNOWN").toString
 
+  def mapSparkFiltersToCypher(filter: Filter, container: org.neo4j.cypherdsl.core.PropertyContainer, attributeAlias: Option[String] = None): Condition = {
+    filter match {
+      case eqns: EqualNullSafe => container.property(attributeAlias.getOrElse(eqns.attribute)).isEqualTo(Cypher.literalOf(eqns.value))
```
jexp, line 222: todo parameters instead of literals, can be in a later PR

jexp, line 222: coalesce around the value?

src/test/scala/org/neo4j/spark/DataSourceReaderTSE.scala
```diff
@@ -330,7 +327,259 @@ class DataSourceReaderTSE extends SparkConnectorScalaBaseTSE {
...
+  def testReadNodeWithGtFilter(): Unit = {
+    val df: DataFrame = initTest(s"""
+     CREATE (p1:Person {age: 19}),
+      (p2:Person {age: 20}),
+      (p3:Person {age: 21})
+     """)
+
+    val result = df.where("age > 20").collectAsList()
+
+    assertEquals(1, result.size())
```
jexp, line 365: these tests should probably check the actual entities are correct in the results, not just the count

src/test/scala/org/neo4j/spark/DataSourceReaderTSE.scala
```diff
@@ -330,7 +327,259 @@ class DataSourceReaderTSE extends SparkConnectorScalaBaseTSE {
...
+        })
+
+    val df = ss.read.format(classOf[DataSource].getName)
+      .option("url", SparkConnectorScalaSuiteIT.server.getBoltUrl)
+      .option("relationship", "KNOWS")
+      .option("relationship.source.labels", "Person")
+      .option("relationship.target.labels", "Person")
+      .load()
+
+    assertEquals(1, df.filter("`<source>`.`id` = '10' AND `<target>`.`id` = '1'").collectAsList().size())
```
jexp, line 549: df can do count without collectToList

src/test/scala/org/neo4j/spark/DataSourceReaderTSE.scala
```diff
@@ -330,7 +327,259 @@ class DataSourceReaderTSE extends SparkConnectorScalaBaseTSE {
...
+
+    val df = ss.read.format(classOf[DataSource].getName)
+      .option("url", SparkConnectorScalaSuiteIT.server.getBoltUrl)
+      .option("relationship", "KNOWS")
+      .option("relationship.nodes.map", "false")
+      .option("relationship.source.labels", "Person")
+      .option("relationship.target.labels", "Person")
+      .load()
+
+    assertEquals(1, df.filter("`source.id` = 10 AND `target.id` = 1").collectAsList().size())
```
jexp, line 578: the syntax is different from before where it was `<source>.<id>` ?
why? we should allow only one (simpler one)

doc/docs/modules/ROOT/pages/quickstart.adoc
```diff
@@ -371,9 +385,38 @@ otherwise if `false`:
...
+
+If `relationship.node.map` is set to **false**
+
+* ``\`source.[property]` `` for the source node properties
+* ``\`rel.[property]` `` for the relation property
+* ``\`target.[property]` `` for the target node property
+
+If `relationship.node.map` is set to **true**
+
+* ``\`<source>`.\`[property]` `` for the source node map properties
```
jexp, line 400: why?

src/test/scala/org/neo4j/spark/DataSourceReaderWithApocTSE.scala
```diff
@@ -326,6 +326,258 @@ class DataSourceReaderWithApocTSE extends SparkConnectorScalaBaseWithApocTSE {
...
+  }
+
+  @Test
+  def testReadNodeWithNotEqualToFilter(): Unit = {
+    val df: DataFrame = initTest(s"""
+     CREATE (p1:Person {name: 'John Doe'}),
+      (p2:Person {name: 'Jane Doe'})
+     """)
+
+    val result = df.select("name").where("NOT name = 'John Doe'").collectAsList()
```
jexp, line 349: is there also a not operator on the predicate API?

src/test/scala/org/neo4j/spark/util/Neo4jImplicitsTest.scala
```diff
@@ -43,5 +43,27 @@ class Neo4jImplicitsTest {
...
+
+    // when
+    val attribute = filter.getAttribute
+
+    // then
+    assertTrue(attribute.isDefined)
+  }
 
+  @Test
+  def `should not return attribute if filter doesn't have it` {
```
jexp, line 59: not sure what this test means.

src/test/scala/org/neo4j/spark/util/Neo4jImplicitsTest.scala
```diff
@@ -43,5 +43,27 @@ class Neo4jImplicitsTest {
     assertEquals(value, actual)
   }
 
+  @Test
```
jexp, line 46: should the other implicits also be tested?

src/test/scala/org/neo4j/spark/Neo4jQueryServiceTest.scala
```diff
@@ -13,7 +14,7 @@ class Neo4jQueryServiceTest {
     options.put(QueryType.LABELS.toString.toLowerCase, "Person")
     val neo4jOptions: Neo4jOptions = new Neo4jOptions(options)
 
-    val query: String = new Neo4jQueryService(neo4jOptions, new Neo4jQueryReadStrategy()).createQuery()
+    val query: String = new Neo4jQueryService(neo4jOptions, new Neo4jQueryReadStrategy(Array[Filter]())).createQuery()
```
jexp, line 17: shouldn't it have a default value of an empty array instead?

src/test/scala/org/neo4j/spark/Neo4jQueryServiceTest.scala
```diff
@@ -25,8 +26,24 @@ class Neo4jQueryServiceTest {
...
+    options.put(QueryType.LABELS.toString.toLowerCase, "Person")
+    val neo4jOptions: Neo4jOptions = new Neo4jOptions(options)
+
+    val filters: Array[Filter] = Array[Filter](
+      EqualTo("name", "John Doe")
+    )
+
+    val query: String = new Neo4jQueryService(neo4jOptions, new Neo4jQueryReadStrategy(filters)).createQuery()
+
+    assertEquals("MATCH (n:`Person`) WHERE n.name = 'John Doe' RETURN n", query)
```
jexp, line 47: please also test the generation of a few more queries

- and/or/not combos (esp. with parentheses around OR groups)
- source/target/rel filters both with map-mode and without
- multi label?
- label filters?

jexp:
Are there also other predicates possible and functions possible? like `toLower` or `abs()` or `WHERE n:Label` ?

doc/docs/modules/ROOT/pages/quickstart.adoc
```diff
@@ -177,6 +177,11 @@ every single node property as column prefixed by `source` or `target`
 |`sample`
 |No
 
+|`pushdown.filters.enabled`
```
moxious, line 180: just out of curiosity, why would I ever not want this enabled?   What situation would mean I should disable it?

doc/docs/modules/ROOT/pages/quickstart.adoc
utnaf: If I got your question right: this is used to filter the source, rel, or target columns when are returned as a map (using the `relationship.node.map` == true). This is the syntax used to access properties inside said maps.

doc/docs/modules/ROOT/pages/quickstart.adoc
utnaf: Maybe some users prefer to leave the filtering to Spark because, for instance, Spark would be faster in filtering some properties that might not have been indexed on Neo4j. 

Also, if we have a bug on the filtering, this could be a nice patch 😅

src/main/scala/org/neo4j/spark/util/Neo4jImplicits.scala
utnaf: > what about list data type (IN)

these are already handled (see line 72 and [Neo4jUtil:228](https://github.com/neo4j-contrib/neo4j-spark-connector/pull/162/files/ede9ef43dceeab9fbb86e7ceee96f45a6f6824aa#diff-f4be0e8f0dc4f5a8b2cd064a79a7461bR228)

> boolean values (no comparion where f.flag )
Spark doesn't flag boolean filtering "Neo4j style" but needs to be explicit: `f.flag = true`

src/test/scala/org/neo4j/spark/DataSourceReaderWithApocTSE.scala
utnaf: We added a test with the `!=` operator

utnaf:
> Are there also other predicates possible and functions possible? like `toLower` or `abs()` or `WHERE n:Label` ?

for functions: you should use Spark functions, since they're not converted to Filters and pushed down to Neo4j. At the moment any way to handle them.

for the labels: they are implicitly defined when we use the `.option("labels", ":Person:Customer")`, same thing applies to relationship
```
.option("relationship", "BOUGHT")
.option("relationship.source.labels", ":Person:Customer")
.option("relationship.target.labels", ":Product")
```

doc/docs/modules/ROOT/pages/quickstart.adoc
jexp: but why needs to be a different syntax there, can't we figure that out internally? and use the appropriate adjusted filter. Just trying to understand.

src/test/scala/org/neo4j/spark/DataSourceReaderWithApocTSE.scala
jexp: I meant on the spark filter API i.e. can I do .not().where() ?

src/test/scala/org/neo4j/spark/Neo4jQueryServiceTest.scala
```diff
@@ -25,8 +26,143 @@ class Neo4jQueryServiceTest {
...
+
+    val filters: Array[Filter] = Array[Filter](
+      Or(EqualTo("source.name", "John Doe"), EqualTo("target.name", "John Doe"))
+    )
+
+    val query: String = new Neo4jQueryService(neo4jOptions, new Neo4jQueryReadStrategy(filters)).createQuery()
+
+    assertEquals("MATCH (source:`Person`) " +
+      "MATCH (target:`Person`) " +
+      "MATCH (source)-[rel:`KNOWS`]->(target) WHERE (source.name = coalesce('John Doe', '') OR target.name = coalesce('John Doe', '')) RETURN source, rel, target", query)
```
jexp, line 89: hmm not sure about the semantics of EqualTo in spark?
b.c. now we also get all the records returned that have `name=''` ? not sure if that's intended.

jexp: Actually I'm not sure we have to distinguish between null and not null equalto?

conker84:
> I don't think the coalesce approach works here as it would return wrong data.
> Do you have a pointer to the "null" semantics for the spark pushdown filters?
> Otherwise I would just use regular cypher expressions and ignore the null bit.
> 
> If you want to specifiy `x IS NULL OR x = 'foo'` I would generate exactly that but that will add a penalty to the cypher query performance.
> 
> Looks good otherwise.

conker84:
@jexp I update the PR as you requested please lemme know what are your thoughts about.

src/main/scala/org/neo4j/spark/util/Neo4jUtil.scala
```diff
@@ -205,15 +206,44 @@ object Neo4jUtil {
...
+      case lte: LessThanOrEqual => container.property(attributeAlias.getOrElse(lte.attribute))
+        .lte(Cypher.literalOf(lte.value))
+      case in: In => {
+        val values = in.values.map(Cypher.literalOf)
+        container.property(attributeAlias.getOrElse(in.attribute)).in(Cypher.literalOf(values.toIterable.asJava))
+      }
+      case notNull: IsNotNull => container.property(attributeAlias.getOrElse(notNull.attribute)).isNotNull
+      case isNull: IsNull => container.property(attributeAlias.getOrElse(isNull.attribute)).isNull
+      case startWith: StringStartsWith => container.property(attributeAlias.getOrElse(startWith.attribute))
+        .startsWith(Functions.coalesce(Cypher.literalOf(startWith.value), Cypher.literalOf("")))
```
jexp, line 241: I thought you wanted to remove the coalesce?

jexp: There might be missing test that would have spotted them. :)

src/test/scala/org/neo4j/spark/Neo4jQueryServiceTest.scala
```diff
@@ -25,8 +26,160 @@ class Neo4jQueryServiceTest {
...
+
+  @Test
+  def testNodeFilterEqualNullSafe(): Unit = {
+    val options: java.util.Map[String, String] = new java.util.HashMap[String, String]()
+    options.put(Neo4jOptions.URL, "bolt://localhost")
+    options.put(QueryType.LABELS.toString.toLowerCase, "Person")
+    val neo4jOptions: Neo4jOptions = new Neo4jOptions(options)
+
+    val filters: Array[Filter] = Array[Filter](
+      EqualNullSafe("name", "John Doe"),
```
jexp, line 58: where are the semantics of EqualNullSafe defined?

src/main/scala/org/neo4j/spark/util/Neo4jUtil.scala
jexp: @conker84 ping

conker84:
@jexp PR updated

doc/docs/modules/ROOT/pages/quickstart.adoc
conker84: We have two ways to represent a relationship:
* with `relationship.node.map=false` => we have the nodes and relationship flattened so every node has the namespace `source/target` and the relationship has `rel`. This is the most descriptive representation;
* with `relationship.node.map=true` => we have the nodes represented as map, in order to compact the table representation.

doc/docs/modules/ROOT/pages/quickstart.adoc
conker84: done at line 213

src/test/scala/org/neo4j/spark/DataSourceReaderWithApocTSE.scala
conker84: No you can't

PR #164, opened 2020-08-13
fixes #163: Add support to partitioning

### LINKED ISSUE #163: Add support to partitioning
Add support to partitioning in order to parallelize the import of the data from Neo4j

conker84:
fixes #163 

We support partitions via skip/limit queries. In order to do so, we compute the total count, for nodes/relationships we leverage the Neo4j count store in order to retrieve the total count about the nodes/relationships we're trying pulling off, for the query we have two possible approaches:

* Compute a count over the query that we're using
* Compute a count over a second *optimized* query that leverages indexes, in this case, you can pass
it via the `.option("query.count", "<your cypher query>")` the query must always return only
one field named `count` which is the result of the count. ie.:

```cypher
MATCH (p:Person)-[r:BOUGHT]->(pr:Product)
WHERE pr.name = 'An Awesome Product'
RETUNR count(p) AS count
```

### DIFF
--- a/doc/docs/modules/ROOT/pages/quickstart.adoc
+++ b/doc/docs/modules/ROOT/pages/quickstart.adoc
@@ -140,6 +140,21 @@ will route the requests to the followers, otherwise to the leader.
 |``
 |Yes^*^
 
+|`query.count`
+|Query count, used only in combination with `query` option, it's a query that returns a `count`
+field like the following.
+``
+MATCH (p:Person)-[r:BOUGHT]->(pr:Product)
+WHERE pr.name = 'An Awesome Product'
+RETURN count(p) AS count
+``
+
+or *a simple number* that represents the count.
+Consider that the number passed by this value represent the volume of the data pulled of Neo4j, so please
+use it carefully.
+|``
+|No
+
 |`labels`
 |List of labels separated by `:`. First label is the primary label
 |``
@@ -182,6 +197,13 @@ every single node property as column prefixed by `source` or `target`
 |`true`
 |No
 
+|`partitions`
+|This defines the parallelization level while pulling data from Neo4j.
+N.b. As More parallelization does not mean more performances so please tune wisely in according to
+your Neo4j installation.
+|`1`
+|No
+
 |===
 
 ^*^ Just one of the options can be specified.
@@ -193,6 +215,61 @@ Reading data from a Neo4j Database can be done in 3 ways:
  * with a Cypher query
  * with a set of node Labels 
  * by specifying a relationship
+
+=== Considerations about partitioning
+
+While we're trying to pull off the data we offer the possibility to partition the extraction in order
+parallelizing it.
+
+Please consider the following job:
+```scala
+val df = spark.read.format("org.neo4j.spark.DataSource")
+        .option("url", "bolt://localhost:7687")
+        .option("labels", "Person")
+        .option("partitions", "5")
+        .load()
+```
+
+This means that if the total count of the nodes with label `Person` into Neo4j is 100 we are creating 5
+partitions and each one will manage 20 records via the `skip/limit` query.
+
+Partitioning the dataset makes sense only if you're dealing with a big dataset (>= 10M of records).
+
+==== How we parallelize the query execution
+
+Considering that we have three options
+
+1. Node extraction
+2. Relationship extraction
+3. Query extraction
+
+We adopt generally provide a general count on what you're trying to pull of and add build
+a query with the skip/limit approach over each partition.
+
+So for a dataset of 100 nodes (Person) with a partition size of 5 we'll generate these queries (one for partition):
+
+```
+MATCH (p:Person) RETURN p SKIP 0 LIMIT 20
+MATCH (p:Person) RETURN p SKIP 20 LIMIT 20
+MATCH (p:Person) RETURN p SKIP 40 LIMIT 20
+MATCH (p:Person) RETURN p SKIP 60 LIMIT 20
+MATCH (p:Person) RETURN p SKIP 80 LIMIT 20
+```
+
+While for (1) and (2) we leverage the Neo4j count store in order to retrieve the total count
+about the nodes/relationships we're trying pulling off, for the (3) we have two possible approaches:
+
+* Compute a count over the query that we're using
+* Compute a count over a second *optimized* query that leverages indexes, in this case you can pass
+it via the `.option("query.count", "<your cypher query>")` the query must always return only
+one field named `count` which is the result of the count. ie.:
+
+```cypher
+MATCH (p:Person)-[r:BOUGHT]->(pr:Product)
+WHERE pr.name = 'An Awesome Product'
+RETURN count(p) AS count
+```
+
 
 === Considerations on the schema
 
--- a/src/main/scala/org/neo4j/spark/reader/Neo4jDataSourceReader.scala
+++ b/src/main/scala/org/neo4j/spark/reader/Neo4jDataSourceReader.scala
@@ -8,7 +8,7 @@ import org.apache.spark.sql.sources.v2.DataSourceOptions
 import org.apache.spark.sql.sources.v2.reader.{DataSourceReader, InputPartition, SupportsPushDownFilters, SupportsPushDownRequiredColumns}
 import org.apache.spark.sql.types.StructType
 import org.neo4j.spark.{DriverCache, Neo4jOptions}
-import org.neo4j.spark.service.SchemaService
+import org.neo4j.spark.service.{PartitionSkipLimit, SchemaService}
 import org.neo4j.spark.util.Validations
 
 import scala.collection.JavaConverters._
@@ -25,12 +25,12 @@ class Neo4jDataSourceReader(private val options: DataSourceOptions, private val 
   override def readSchema(): StructType = callSchemaService { schemaService => schemaService
     .struct() }
 
-  private def callSchemaService[T](lambda: SchemaService => T): T = {
+  private def callSchemaService[T](function: SchemaService => T): T = {
     val driverCache = new DriverCache(neo4jOptions.connection, jobId)
     val schemaService = new SchemaService(neo4jOptions, driverCache)
     var hasError = false
     try {
-      lambda(schemaService)
+      function(schemaService)
     } catch {
       case e: Throwable => {
         hasError = true
@@ -44,20 +44,21 @@ class Neo4jDataSourceReader(private val options: DataSourceOptions, private val 
     }
   }
 
-  private def getPartitions(): Seq[(Long, Long)] = callSchemaService { schemaService => schemaService
-    .skipLimitFromPartition() }
-
   override def planInputPartitions: util.ArrayList[InputPartition[InternalRow]] = {
+    // we retrieve the schema in order to parse the data correctly
     val schema = readSchema()
-    val partitionSkipLimit = getPartitions()
-    val neo4jPartitions = if (partitionSkipLimit.isEmpty) {
-      Seq(new Neo4jInputPartitionReader(neo4jOptions, filters, schema, jobId))
-    } else {
-      partitionSkipLimit.zipWithIndex
-        .map(triple => new Neo4jInputPartitionReader(neo4jOptions, filters, schema, jobId, triple._2,
-          triple._1._1, triple._1._2))
-    }
-    new java.util.ArrayList[InputPartition[InternalRow]](neo4jPartitions.asJava)
+    val neo4jPartitions: Seq[Neo4jInputPartitionReader] = createPartitions(schema)
+    new util.ArrayList[InputPartition[InternalRow]](neo4jPartitions.asJava)
+  }
+
+  private def createPartitions(schema: StructType) = {
+    // we get the skip/limit for each partition
+    val partitionSkipLimitList = callSchemaService { schemaService => schemaService
+      .skipLimitFromPartition() }
+    // we generate a partition for each element
+    partitionSkipLimitList
+      .map(partitionSkipLimit => new Neo4jInputPartitionReader(neo4jOptions, filters, schema, jobId,
+        partitionSkipLimit))
   }
 
   override def pushFilters(filtersArray: Array[Filter]): Array[Filter] = {
--- a/src/main/scala/org/neo4j/spark/reader/Neo4jDataSourceReader.scala
+++ b/src/main/scala/org/neo4j/spark/reader/Neo4jDataSourceReader.scala
@@ -7,9 +7,11 @@ import org.apache.spark.sql.sources.Filter
 import org.apache.spark.sql.sources.v2.DataSourceOptions
 import org.apache.spark.sql.sources.v2.reader.{DataSourceReader, InputPartition, SupportsPushDownFilters, SupportsPushDownRequiredColumns}
 import org.apache.spark.sql.types.StructType
-import org.neo4j.spark.Neo4jOptions
+import org.neo4j.spark.{DriverCache, Neo4jOptions}
 import org.neo4j.spark.service.SchemaService
 import org.neo4j.spark.util.Validations
+
+import scala.collection.JavaConverters._
 
 class Neo4jDataSourceReader(private val options: DataSourceOptions, private val jobId: String) extends DataSourceReader
   with SupportsPushDownFilters {
@@ -19,20 +21,43 @@ class Neo4jDataSourceReader(private val options: DataSourceOptions, private val 
   private val neo4jOptions: Neo4jOptions = new Neo4jOptions(options.asMap())
     .validate(options => Validations.read(options, jobId))
 
-  override def readSchema(): StructType = {
-    val schemaService = new SchemaService(neo4jOptions, jobId)
+
+  override def readSchema(): StructType = callSchemaService { schemaService => schemaService
+    .struct() }
+
+  private def callSchemaService[T](lambda: SchemaService => T): T = {
+    val driverCache = new DriverCache(neo4jOptions.connection, jobId)
+    val schemaService = new SchemaService(neo4jOptions, driverCache)
+    var hasError = false
     try {
-      schemaService.struct()
+      lambda(schemaService)
+    } catch {
+      case e: Throwable => {
+        hasError = true
+        throw e
+      }
     } finally {
       schemaService.close()
+      if (hasError) {
+        driverCache.close()
+      }
     }
   }
 
+  private def getPartitions(): Seq[(Long, Long)] = callSchemaService { schemaService => schemaService
+    .skipLimitFromPartition() }
+
   override def planInputPartitions: util.ArrayList[InputPartition[InternalRow]] = {
     val schema = readSchema()
-    val factoryList = new java.util.ArrayList[InputPartition[InternalRow]]
-    factoryList.add(new Neo4jInputPartitionReader(neo4jOptions, filters, schema, jobId))
-    factoryList
+    val partitionSkipLimit = getPartitions()
+    val neo4jPartitions = if (partitionSkipLimit.isEmpty) {
+      Seq(new Neo4jInputPartitionReader(neo4jOptions, filters, schema, jobId))
+    } else {
+      partitionSkipLimit.zipWithIndex
+        .map(triple => new Neo4jInputPartitionReader(neo4jOptions, filters, schema, jobId, triple._2,
+          triple._1._1, triple._1._2))
+    }
+    new java.util.ArrayList[InputPartition[InternalRow]](neo4jPartitions.asJava)
   }
 
   override def pushFilters(filtersArray: Array[Filter]): Array[Filter] = {
--- a/src/main/scala/org/neo4j/spark/reader/Neo4jInputPartitionReader.scala
+++ b/src/main/scala/org/neo4j/spark/reader/Neo4jInputPartitionReader.scala
@@ -6,7 +6,7 @@ import org.apache.spark.sql.sources.Filter
 import org.apache.spark.sql.sources.v2.reader.{InputPartition, InputPartitionReader}
 import org.apache.spark.sql.types.StructType
 import org.neo4j.driver.{Record, Session, Transaction}
-import org.neo4j.spark.service.{MappingService, Neo4jQueryReadStrategy, Neo4jQueryService, Neo4jReadMappingStrategy}
+import org.neo4j.spark.service.{MappingService, Neo4jQueryReadStrategy, Neo4jQueryService, Neo4jReadMappingStrategy, PartitionSkipLimit}
 import org.neo4j.spark.util.Neo4jUtil
 import org.neo4j.spark.{DriverCache, Neo4jOptions}
 
@@ -16,9 +16,7 @@ class Neo4jInputPartitionReader(private val options: Neo4jOptions,
                                 private val filters: Array[Filter],
                                 private val schema: StructType,
                                 private val jobId: String,
-                                private val partitionId: Int = 0,
-                                private val skip: Long = -1L,
-                                private val limit: Long = -1L) extends InputPartition[InternalRow]
+                                private val partitionSkipLimit: PartitionSkipLimit) extends InputPartition[InternalRow]
   with InputPartitionReader[InternalRow]
   with Logging {
 
@@ -26,16 +24,15 @@ class Neo4jInputPartitionReader(private val options: Neo4jOptions,
   private var session: Session = _
   private var transaction: Transaction = _
   private val driverCache: DriverCache = new DriverCache(options.connection,
-    if (partitionId > 0) s"$jobId-$partitionId" else jobId)
+    if (partitionSkipLimit.partitionNumber > 0) s"$jobId-${partitionSkipLimit.partitionNumber}" else jobId)
 
-  private val query: String = new Neo4jQueryService(options, new Neo4jQueryReadStrategy(filters))
+  private val query: String = new Neo4jQueryService(options, new Neo4jQueryReadStrategy(filters, partitionSkipLimit))
     .createQuery()
-    .concat(if (skip != -1 && limit != -1) s" SKIP $$skip LIMIT $$limit" else "")
 
   private val mappingService = new MappingService(new Neo4jReadMappingStrategy(options), options)
 
   override def createPartitionReader(): InputPartitionReader[InternalRow] = new Neo4jInputPartitionReader(options, filters, schema,
-    jobId, partitionId, skip, limit)
+    jobId, partitionSkipLimit)
 
   def next: Boolean = {
     if (result == null) {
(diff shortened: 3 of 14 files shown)


doc/docs/modules/ROOT/pages/quickstart.adoc
```diff
@@ -177,6 +188,13 @@ every single node property as column prefixed by `source` or `target`
 |`sample`
 |No
 
+|`partitions`
+|This defines the parallelization level while pulling data from Neo4j.
+N.b. As More parallelization does not mean more performances so please tune wisely in according to
```
moxious, line 193: "As more parallelization does not mean better performance, choose this setting wisely"

Note the interaction with "count".   I think what you mean is that if the count query returns 100 and partitions=5, you're going to be getting 5 partitions of 20 records, using skip/limit.   You should just say that.

doc/docs/modules/ROOT/pages/quickstart.adoc
```diff
@@ -189,6 +207,36 @@ Reading data from a Neo4j Database can be done in 3 ways:
...
+
+* Compute a count over the query that we're using
+* Compute a count over a second *optimized* query that leverages indexes, in this case you can pass
+it via the `.option("query.count", "<your cypher query>")` the query must always return only
+one field named `count` which is the result of the count. ie.:
+
+```cypher
+MATCH (p:Person)-[r:BOUGHT]->(pr:Product)
+WHERE pr.name = 'An Awesome Product'
+RETUNR count(p) AS count
```
moxious, line 236: typo return

doc/docs/modules/ROOT/pages/quickstart.adoc
```diff
@@ -189,6 +207,36 @@ Reading data from a Neo4j Database can be done in 3 ways:
...
+
+==== How we parallelize the query execution
+
+Considering that we have three options
+
+1. Node extraction
+2. Relationship extraction
+3. Query extraction
+
+We adopt generally provide a general count on what you're trying to pull of and add build
```
moxious, line 223: provide an example here, this isn't very clear.  For example, I'm loading just Person nodes.  I set partitions=5.  This turns into:

```
MATCH (p:Person) RETURN p LIMIT 20 SKIP 0
MATCH (p:Person) RETURN p LIMIT 20 SKIP 20
...
```

doc/docs/modules/ROOT/pages/quickstart.adoc
```diff
@@ -189,6 +207,36 @@ Reading data from a Neo4j Database can be done in 3 ways:
...
+a query with the skip/limit approach over each partition.
+While for (1) and (2) we leverage the Neo4j count store in order to retrieve the total count
+about the nodes/relationships we're trying pulling off, for the (3) we have two possible approaches:
+
+* Compute a count over the query that we're using
```
moxious, line 228: How would this be done in the general query case?  Or it can't be done, and using more than > 1 partition with a query where no count is specified will fail?

src/main/scala/org/neo4j/spark/service/SchemaService.scala
```diff
@@ -158,6 +163,104 @@ class SchemaService(private val options: Neo4jOptions, private val jobId: String
...
+                             |""".stripMargin
+              val count = session.run(query).single().get("count")
+              if (count.isNull) 0L else count.asLong()
+            })
+            .min
+        }
+      }
+  }
+
+  def countForRelationship(): Long = try {
```
moxious, line 200: These count methods, please think about failure methods.  In some of these APOC calls you are doing you are coupling down to the string format that APOC returns in a particular field.  Remembering that APOC will get put into product at some point, and the string format of return columns isn't promised, this code has some potential to break at some point in the future due to a new published APOC, just as an example of what can go wrong.

Suppose the user specifies paralleism of 5, and *everything goes wrong* with the count feature, and there's no way to obtain a valid count.   What happens then?    My vote would be for a debug message, plus reverting to reading into one partition (with no need for counts)

moxious: there's the extra possibility that the user just regular screws up the count query.   Say I specify the count query like this:

```
MATCH (p:Person)-[:BOUGHT]->(pr:Product)
RETURN count(pr) - 10 as count
```

Doesn't really matter how or why, but the count is wrong.

In this case, you might try to fetch one extra "empty" partition.  If the user says 5 partitions, try to fetch 6, where the 6th partition has one record (SKIP whatever LIMIT 1).   If you get nothing, that's good.  If you get an actual record here, this is a warning that the user has messed up the partition.   Another warning is if any of your partitions end up empty, then chances are good the user messed up the count query.

src/test/scala/org/neo4j/spark/DataSourceReaderTSE.scala
```diff
@@ -351,6 +348,110 @@ class DataSourceReaderTSE extends SparkConnectorScalaBaseTSE {
...
+        """
+          |MATCH (p:Person)-[r:BOUGHT]->(pr:Product)
+          |RETURN p.name AS person, pr.name AS product, r.quantity AS quantity""".stripMargin)
+      .option("partitions", "5")
+      .load()
+
+    assertEquals(5, partitionedDf.rdd.getNumPartitions)
+    assertEquals(150, partitionedDf.collect().map(_.getAs[Long]("person")).toSet.size)
+
+    val partitionedQueryCountDf = ss.read.format(classOf[DataSource].getName)
```
moxious, line 439: Add tests:

- What happens when query.count is invalid, or fails to return a "count" field, or returns a "count" field that has nonsense?

jexp:
Actually the optimized version should be:

```
MATCH (pr:Product)
WHERE pr.name = 'An Awesome Product'
RETURN size(()-[:BOUGHT]->(pr)) AS count
```
which uses get-degree

doc/docs/modules/ROOT/pages/quickstart.adoc
```diff
@@ -140,6 +140,17 @@ will route the requests to the followers, otherwise to the leader.
 |``
 |Yes^*^
 
+|`query.count`
+|Query count, used only in combination with `query` option, it's a query that returns a `count`
+field like the following
+```
+MATCH (p:Person)-[r:BOUGHT]->(pr:Product)
```
jexp, line 147: see my optimized version in the other comment.

jexp: also typo in RETURN

doc/docs/modules/ROOT/pages/quickstart.adoc
jexp: In theory we could also leverage cursors:

- given there is a unqiue constraint
- we can return the data ordered (using index backed order by) with a limit
- then we can take the last value returned and add an `WHERE n.id > $cursor` which then would only query from that value
- pure skip limit queries need to re-execute and throw away, so they get slower over time

doc/docs/modules/ROOT/pages/quickstart.adoc
jexp: We should also mention that it only makes sense when you're pulling more than X (where X can be 1M or 10M)

doc/docs/modules/ROOT/pages/quickstart.adoc
jexp: Can I also opt out of the count query and just provide a literal value if I know how much data is there?
That would be really useful.

src/main/scala/org/neo4j/spark/service/SchemaService.scala
```diff
@@ -109,8 +106,16 @@ class SchemaService(private val options: Neo4jOptions, private val jobId: String
     }
 
     structFields ++= (try {
-        val query = "CALL apoc.meta.relTypeProperties({ includeRels: $rels })"
-        val params = Map[String, AnyRef]("rels" -> Seq(options.relationshipMetadata.relationshipType).asJava)
+        val query =
+          """CALL apoc.meta.relTypeProperties({ includeRels: $rels }) YIELD sourceNodeLabels, targetNodeLabels,
+            | propertyName, propertyTypes
+            |WITH *
+            |WHERE sourceNodeLabels = $sourceLabels AND targetNodeLabels = $targetLabels
```
jexp, line 113: shouldn't we rather pass this to the proc? seems a big waste. If it's based on apoc.meta.* it should support include and exclude functionality.

src/main/scala/org/neo4j/spark/service/SchemaService.scala
```diff
@@ -158,6 +163,104 @@ class SchemaService(private val options: Neo4jOptions, private val jobId: String
...
+    map.filterKeys(k => options.nodeMetadata.labels.contains(k))
+      .values.min
+  } catch {
+    case e: ClientException =>
+      e.code match {
+        case "Neo.ClientError.Procedure.ProcedureNotFound" => {
+          // TODO get back to Cypher DSL
+          options.nodeMetadata.labels
+            .map(label => {
+              val query = s"""MATCH (${Neo4jUtil.NODE_ALIAS}:$label)
```
jexp, line 189: you can do a union query that uses 

`MATCH (:`$label`) RETURN { type: '$label', count:count(*) } as counts UNION ALL ...`

src/main/scala/org/neo4j/spark/service/SchemaService.scala
jexp: just allow to specifiy a literal count value in the property

jexp: It's always good to overfetch for stuff like that in SDN we used to pull pages with pagesize+1 so we could say "there is a next page"

src/main/scala/org/neo4j/spark/service/SchemaService.scala
```diff
@@ -158,6 +163,104 @@ class SchemaService(private val options: Neo4jOptions, private val jobId: String
...
+                 |RETURN count(${Neo4jUtil.RELATIONSHIP_ALIAS}) AS count
+                 |""".stripMargin).single().get("count").asLong())
+            .min
+          val minFromTarget = options.relationshipMetadata.target.labels
+            .map(label => session.run(s"""MATCH ()-[${Neo4jUtil.RELATIONSHIP_ALIAS}:${options.relationshipMetadata.relationshipType}]->(:$label)
+                                         |RETURN count(${Neo4jUtil.RELATIONSHIP_ALIAS}) AS count
+                                         |""".stripMargin).single().get("count"))
+            .map(count => if (count.isNull) 0L else count.asLong())
+            .min
+          Math.min(minFromSource, minFromTarget)
```
jexp, line 233: why min and not max?

src/main/scala/org/neo4j/spark/service/SchemaService.scala
```diff
@@ -158,6 +163,104 @@ class SchemaService(private val options: Neo4jOptions, private val jobId: String
...
+        }
+      }
+  }
+
+  def countForQuery(): Long = {
+    val query = if (options.queryMetadata.queryCount.trim.nonEmpty) {
+      options.queryMetadata.queryCount
+    } else {
+      s"""CALL { ${options.query.value} }
```
jexp, line 242: clever use of subquery

src/main/scala/org/neo4j/spark/service/SchemaService.scala
conker84: Just to be sure should we over-fetch always or only if the user provides the query count right?

conker84: > just allow to specifiy a literal count value in the property

the user can define a query like this `RETURN 50 AS count` is it fine? Or we allow to simply pass `.option("query.count", "50")`?

src/test/scala/org/neo4j/spark/DataSourceReaderTSE.scala
conker84: We provided a check if was a READ_ONLY query, I added a check about the returned identifier, that must be only `count`.
What do you mean nonsense? About type? That will be a runtime error because in my knowledge

src/main/scala/org/neo4j/spark/service/SchemaService.scala
conker84: I added `.option("query.count", "50")` as valid option

src/main/scala/org/neo4j/spark/service/SchemaService.scala
conker84: because I use the most selective, if we have this use use-case:

```scala
      .option("relationship", "BOUGHT")
      .option("relationship.nodes.map", "false")
      .option("relationship.source.labels", ":Person:RegisteredCustomer")
      .option("relationship.target.labels", ":Product")
      .load()
```

This will get three counts:

- `(:Person)-[:BOUGHT]->()` => 1000 rows
- `(:RegisteredCustomer)-[:BOUGHT]->()` => 250 rows
- `()-[:BOUGHT]->(:Product)` => 1000

we use the most selective.

src/main/scala/org/neo4j/spark/service/SchemaService.scala
conker84: Unfortunately at this moment, the procedure does not take those parameters as configuration params

doc/docs/modules/ROOT/pages/quickstart.adoc
conker84: Done

src/test/scala/org/neo4j/spark/DataSourceReaderTSE.scala
conker84: github seems trucated my comment 😥
... in my knowledge there is no way to get the return type from the explain

src/main/scala/org/neo4j/spark/reader/Neo4jInputPartitionReader.scala
```diff
@@ -15,27 +15,37 @@ import scala.collection.JavaConverters._
...
   private var session: Session = _
   private var transaction: Transaction = _
-  private val driverCache: DriverCache = new DriverCache(options.connection, jobId)
+  private val driverCache: DriverCache = new DriverCache(options.connection,
+    if (partitionId > 0) s"$jobId-$partitionId" else jobId)
 
-  private val query: String = new Neo4jQueryService(options, new Neo4jQueryReadStrategy(filters)).createQuery()
+  private val query: String = new Neo4jQueryService(options, new Neo4jQueryReadStrategy(filters))
+    .createQuery()
+    .concat(if (skip != -1 && limit != -1) s" SKIP $$skip LIMIT $$limit" else "")
```
jexp, line 33: it would be nicer if we had a proper method on createQuery i.e. pass it to the query builder

src/main/scala/org/neo4j/spark/reader/Neo4jInputPartitionReader.scala
```diff
@@ -15,27 +15,37 @@ import scala.collection.JavaConverters._
...
+                                private val skip: Long = -1L,
+                                private val limit: Long = -1L) extends InputPartition[InternalRow]
   with InputPartitionReader[InternalRow]
   with Logging {
 
   private var result: Iterator[Record] = _
   private var session: Session = _
   private var transaction: Transaction = _
-  private val driverCache: DriverCache = new DriverCache(options.connection, jobId)
+  private val driverCache: DriverCache = new DriverCache(options.connection,
```
jexp, line 28: why do we need the partitionid in the driver cache?

src/main/scala/org/neo4j/spark/service/SchemaService.scala
jexp: oh ok, that's really unfortunate.

jexp: But we can change that.

src/main/scala/org/neo4j/spark/service/SchemaService.scala
```diff
@@ -158,11 +168,178 @@ class SchemaService(private val options: Neo4jOptions, private val jobId: String
...
+        .map(label => s"""MATCH ()-[${Neo4jUtil.RELATIONSHIP_ALIAS}:${options.relationshipMetadata.relationshipType}]->(:$label)
+                         |RETURN count(${Neo4jUtil.RELATIONSHIP_ALIAS}) AS count
+                         |""".stripMargin)
+      (sourceQueries ++ targetQueries)
+        .mkString(" UNION ALL ")
+    } else {
+      queryReadStrategy.createStatementForRelationshipCount(options)
+    }
+    log.info(s"Executing the following counting query on Neo4j: $query")
+    session.run(query)
```
jexp, line 209: how do you map them back to the source node / rel-type? or don't you need that?

src/main/scala/org/neo4j/spark/service/SchemaService.scala
```diff
@@ -158,11 +168,178 @@ class SchemaService(private val options: Neo4jOptions, private val jobId: String
...
+      .list()
+      .asScala
+      .map(_.get("count"))
+      .map(count => if (count.isNull) 0L else count.asLong())
+      .min
+  }
+
+  def countForRelationshipWithQuery(filters: Array[Filter]): Long = {
+    val query = if (filters.isEmpty) {
+      val sourceQueries = options.relationshipMetadata.source.labels
```
jexp, line 193: why can't we use apoc.meta.stats here?

src/main/scala/org/neo4j/spark/service/SchemaService.scala
```diff
@@ -158,11 +168,178 @@ class SchemaService(private val options: Neo4jOptions, private val jobId: String
...
+
+  def countForNode(filters: Array[Filter]): Long = try {
+    /*
+     * we try to leverage the count store in order to have the faster response possible
+     * https://neo4j.com/developer/kb/fast-counts-using-the-count-store/
+     * so in this scenario we have some limitations given the fact that we get the min
+     * for the sequence of counts returned
+     */
+    if (filters.isEmpty) {
+      val query = "CALL apoc.meta.stats() yield labels RETURN labels"
```
jexp, line 225: if we only interested in the labels we should probably call db.labels() ?

but not sure how that performs in terms of RBAC

src/main/scala/org/neo4j/spark/service/SchemaService.scala
```diff
@@ -158,11 +168,178 @@ class SchemaService(private val options: Neo4jOptions, private val jobId: String
...
+          countForNodeWithQuery(filters)
+        }
+        case _ => logExceptionForCount(e)
+      }
+    case e: Throwable => logExceptionForCount(e)
+  }
+
+  def countForRelationship(filters: Array[Filter]): Long = try {
+    if (filters.isEmpty) {
+      val query = "CALL apoc.meta.stats() yield relTypes RETURN relTypes"
```
jexp, line 250: same as above db.relationshipTypes() ?

I thought we wanted to use apoc.meta.stats for the actual counts.

src/main/scala/org/neo4j/spark/service/SchemaService.scala
```diff
@@ -158,11 +168,178 @@ class SchemaService(private val options: Neo4jOptions, private val jobId: String
...
+  }
+
+  def skipLimitFromPartition(): Seq[(Long, Long)] = if (options.partitions == 1) {
+    Seq.empty
+  } else {
+    val count: Long = this.count()
+    if (count <= 0) {
+      Seq.empty
+    } else {
+      val partitionSize = Math.ceil(count / options.partitions).toLong
```
jexp, line 315: isn't this a whole number division? so `ceil` has no effect

src/main/scala/org/neo4j/spark/util/Neo4jUtil.scala
```diff
@@ -53,7 +55,8 @@ object Neo4jUtil {
         case _ => autoCloseable.close()
       }
     } catch {
-      case _ => throw new Exception("This exception should be logged") // @todo Log
+      case t: Throwable => if (logger != null) logger
+        .error(s"Cannot close ${autoCloseable.getClass.getSimpleName} because of the following exception:", t)
```
jexp, line 59: this should not be an error, at most a warning

src/main/scala/org/neo4j/spark/reader/Neo4jDataSourceReader.scala
```diff
@@ -19,20 +21,43 @@ class Neo4jDataSourceReader(private val options: DataSourceOptions, private val
...
+  override def readSchema(): StructType = callSchemaService { schemaService => schemaService
+    .struct() }
+
+  private def callSchemaService[T](lambda: SchemaService => T): T = {
+    val driverCache = new DriverCache(neo4jOptions.connection, jobId)
+    val schemaService = new SchemaService(neo4jOptions, driverCache)
+    var hasError = false
     try {
-      schemaService.struct()
+      lambda(schemaService)
```
jexp, line 33: what does lambda() do ?

src/main/scala/org/neo4j/spark/reader/Neo4jDataSourceReader.scala
```diff
@@ -19,20 +21,43 @@ class Neo4jDataSourceReader(private val options: DataSourceOptions, private val
   private val neo4jOptions: Neo4jOptions = new Neo4jOptions(options.asMap())
     .validate(options => Validations.read(options, jobId))
 
-  override def readSchema(): StructType = {
-    val schemaService = new SchemaService(neo4jOptions, jobId)
+
+  override def readSchema(): StructType = callSchemaService { schemaService => schemaService
+    .struct() }
+
+  private def callSchemaService[T](lambda: SchemaService => T): T = {
```
jexp, line 28: supplier ? or factory? instaed of `lambda`?

src/main/scala/org/neo4j/spark/reader/Neo4jDataSourceReader.scala
```diff
@@ -19,20 +21,43 @@ class Neo4jDataSourceReader(private val options: DataSourceOptions, private val
...
+    val driverCache = new DriverCache(neo4jOptions.connection, jobId)
+    val schemaService = new SchemaService(neo4jOptions, driverCache)
+    var hasError = false
     try {
-      schemaService.struct()
+      lambda(schemaService)
+    } catch {
+      case e: Throwable => {
+        hasError = true
```
jexp, line 36: logging?

src/main/scala/org/neo4j/spark/reader/Neo4jDataSourceReader.scala
```diff
@@ -19,20 +21,43 @@ class Neo4jDataSourceReader(private val options: DataSourceOptions, private val
...
   override def planInputPartitions: util.ArrayList[InputPartition[InternalRow]] = {
     val schema = readSchema()
-    val factoryList = new java.util.ArrayList[InputPartition[InternalRow]]
-    factoryList.add(new Neo4jInputPartitionReader(neo4jOptions, filters, schema, jobId))
-    factoryList
+    val partitionSkipLimit = getPartitions()
+    val neo4jPartitions = if (partitionSkipLimit.isEmpty) {
+      Seq(new Neo4jInputPartitionReader(neo4jOptions, filters, schema, jobId))
+    } else {
+      partitionSkipLimit.zipWithIndex
```
jexp, line 56: this code is not very readable, can we use a caseclass instead of the triple?

src/main/scala/org/neo4j/spark/reader/Neo4jDataSourceReader.scala
```diff
@@ -19,20 +21,43 @@ class Neo4jDataSourceReader(private val options: DataSourceOptions, private val
   private val neo4jOptions: Neo4jOptions = new Neo4jOptions(options.asMap())
     .validate(options => Validations.read(options, jobId))
 
-  override def readSchema(): StructType = {
-    val schemaService = new SchemaService(neo4jOptions, jobId)
+
```
jexp, line 24: this whole new section is quite hard to understand / reason about, can you do somethign with the naming of methods/params or add a few comments

jexp (approved):
Please have a look at my comments

src/main/scala/org/neo4j/spark/service/SchemaService.scala
conker84: yes

src/main/scala/org/neo4j/spark/reader/Neo4jInputPartitionReader.scala
conker84: Done

src/main/scala/org/neo4j/spark/reader/Neo4jInputPartitionReader.scala
conker84: Because otherwise in a multi-partition env the first one that closes the driver cache close it for everyone.

src/main/scala/org/neo4j/spark/service/SchemaService.scala
conker84: What do you mean?

src/main/scala/org/neo4j/spark/service/SchemaService.scala
conker84: we use it, this is the fallback in case it's not present, look here:
https://github.com/neo4j-contrib/neo4j-spark-connector/pull/164/files#diff-85bfb2242b03b769ed6b534ff8992f88R255

src/main/scala/org/neo4j/spark/service/SchemaService.scala
conker84: Great catch thank you!!!

src/main/scala/org/neo4j/spark/util/Neo4jUtil.scala
conker84: Done

src/main/scala/org/neo4j/spark/reader/Neo4jDataSourceReader.scala
conker84: I changed the name to `function` if you agree I would leave it with a lambda which keeps the code very simple

src/main/scala/org/neo4j/spark/reader/Neo4jDataSourceReader.scala
conker84: It will logged by Spark, I can add it but it would be redundant

src/main/scala/org/neo4j/spark/reader/Neo4jDataSourceReader.scala
conker84: Done

src/main/scala/org/neo4j/spark/reader/Neo4jDataSourceReader.scala
conker84: done

conker84:
Thank @jexp for the feedback I'm merging this

PR #167, opened 2020-08-25
Add support for both scala 2.11 and scala 2.12

conker84:
Add support for both scala 2.11 and scala 2.12


conker84:
I'm merging this as has really small changes for the CI testing

Issue #159, opened 2020-08-04
Support general scripting or "execute series of commands"


moxious:
So far our current plan is going in the direction of sinking dataframes using this API.  That's good.  A thing that I've already found I'm going to need is a way of just executing an ordered series of cypher queries in a single partition (single-threaded).  Think of this as a way of executing a script.

Before we get into ways to do this -- 

# Use cases for why I want to do this:
* Schema setup / creating indexes prior to a load.  Daily loads might want to begin with deleting a subsection of yesterday's data, for example
* Executing a series of GDS actions (for example creating a virtual graph, and then getting results back from it)
* Cypher scripting from spark (i.e. storing & applying dynamic cypher in other data structures like tables)

This will enhance the ability to treat a single spark job as a big unit of work.  

Question is how best to do this.   Here's a sketch of how I've thought to do this.  Just make scripts dataframes and make the cypher runnable from a column, like so:

## Option 1 - Use "Query"

```
from pyspark.sql import Row
from pyspark.sql.functions import col

script_row = Row("sequence", "cypher", "param1")
my_script = [
    script_row(1, 'CREATE INDEX ON :Thing(id)', None),
    script_row(2, 'CREATE (:Thing { id: event.param1 })', 25)
]
​
script = spark.createDataFrame(data)

script.repartition(1).orderBy('sequence').write
  .format("org.neo4j.spark.DataSource")
  .option("url", "bolt://localhost:7687")
  .option("query", col('cypher'))
  .save()
```

Notice how the query is a DF column not a string.  The query format already gives us the ability to bind the entire row effectively as a set of parameters to the query.  This approach would be problematic with batching though (UNWIND events as event) and we don't want the user to have to remember to set batching=1 every time.   The other approach is a different explicit runmode, like this:

## Option 2 - Introduce "Script" (or something similar)
```
script.write
  .format("org.neo4j.spark.DataSource")
  .option("url", "bolt://localhost:7687")
  .option("script", True)
  .option("script.order", "sequence")
  .option("script.cypher", "cypher")
  .save()
```

This mode would *never* batch and would *always* repartition(1) and execute in the order of the `script.order` column, or else in the order of the first column.  Would always run in "write mode".  Would fail if script.cypher isn't specified.   But like "query", it would always bind the dataframe row to "event", so that by putting extra columns in the dataframe, a user could dynamically pass parameters to themselves.

## Alternatives

[...]

PR #174, opened 2020-09-06
Allow index/constraint creation

conker84:
This addresses part of #159 

The user can set the optimization via `schema.optimization.type` option that takes three values:

* `INDEX`: it creates only indexes on provided nodes
* `CONSTRAINTS`: it creates only indexes on provided nodes
* `QUERY`: it perform a series of schema queries separated by `;`

### DIFF
--- a/doc/docs/modules/ROOT/pages/quickstart_writer.adoc
+++ b/doc/docs/modules/ROOT/pages/quickstart_writer.adoc
@@ -320,3 +320,88 @@ You can specify 3 different modes to use for saving the nodes:
 * `Overwrite`: will perform a `MERGE` on that node
 * `ErrorIfExists`: will perform a `CREATE`
 * `Match`: will perform a `MATCH`
+
+
+=== Schema Optimization Operations
+
+The spark connector supports schema optimization operations via:
+
+* index
+* constraints
+* set of schema queries
+
+that will be executed *before* the import process will start in order to speed-up the import itself.
+
+You can set the optimization via `schema.optimization.type` option that takes three values:
+
+* `INDEX`: it creates only indexes on provided nodes
+* `CONSTRAINTS`: it creates only indexes on provided nodes
+* `QUERY`: it perform a series of schema queries separated by `;`
+
+and it works only when you're merging nodes.
+
+==== Index Creation
+
+Following an example of how to create indexes while you're creating nodes
+
+----
+ds.write
+      .format(classOf[DataSource].getName)
+      .mode(SaveMode.Overwrite)
+      .option("url", SparkConnectorScalaSuiteIT.server.getBoltUrl)
+      .option("labels", ":Person:Customer")
+      .option("node.keys", "surname")
+      .option("schema.optimization.type", "INDEX")
+      .save()
+----
+
+This will create, before the import starts, the following schema query:
+
+----
+CREATE INDEX ON :Person(surname)
+----
+
+*So please into consideration that the first label is used for the index creation*
+
+
+==== Constraint Creation
+
+Following an example of how to create indexes while you're creating nodes
+
+----
+ds.write
+      .format(classOf[DataSource].getName)
+      .mode(SaveMode.Overwrite)
+      .option("url", SparkConnectorScalaSuiteIT.server.getBoltUrl)
+      .option("labels", ":Person:Customer")
+      .option("node.keys", "surname")
+      .option("schema.optimization.type", "CONSTRAINT")
+      .save()
+----
+
+This will create, before the import starts, the following schema query:
+
+----
+CREATE CONSTRAINT ON (p:Person) ASSERT (p.surname) IS UNIQUE
+----
+
+*So please into consideration that the first label is used for the index creation*
+
+==== Index/Constraint Creation via schema query
+
+Following an example of how to create indexes while you're creating nodes
+
+----
+ds.write
+      .format(classOf[DataSource].getName)
+      .mode(SaveMode.Overwrite)
+      .option("url", SparkConnectorScalaSuiteIT.server.getBoltUrl)
+      .option("labels", ":Person:Customer")
+      .option("node.keys", "surname")
+      .option("schema.optimization.type", "QUERY")
+      .option("schema.optimization.query", "CREATE INDEX ON :Person(surname); CREATE CONSTRAINT ON (p:Product) ASSERT (p.name, p.sku) IS NODE KEY")
+      .save()
+----
+
+Before the import starts, the connector will use the schema query that you provided into the
+`schema.optimization.query` option.
--- a/doc/docs/modules/ROOT/pages/quickstart_writer.adoc
+++ b/doc/docs/modules/ROOT/pages/quickstart_writer.adoc
@@ -335,7 +335,7 @@ that will be executed *before* the import process will start in order to speed-u
 You can set the optimization via `schema.optimization.type` option that takes three values:
 
 * `INDEX`: it creates only indexes on provided nodes
-* `CONSTRAINTS`: it creates only indexes on provided nodes
+* `NODE_CONSTRAINTS`: it creates only indexes on provided nodes
 * `QUERY`: it perform a series of schema queries separated by `;`
 
 and it works only when you're merging nodes.
@@ -375,7 +375,7 @@ ds.write
       .option("url", SparkConnectorScalaSuiteIT.server.getBoltUrl)
       .option("labels", ":Person:Customer")
       .option("node.keys", "surname")
-      .option("schema.optimization.type", "CONSTRAINT")
+      .option("schema.optimization.type", "NODE_CONSTRAINTS")
       .save()
 ----
 
--- a/doc/docs/modules/ROOT/pages/quickstart_writer.adoc
+++ b/doc/docs/modules/ROOT/pages/quickstart_writer.adoc
@@ -387,21 +387,39 @@ CREATE CONSTRAINT ON (p:Person) ASSERT (p.surname) IS UNIQUE
 
 *So please into consideration that the first label is used for the index creation*
 
-==== Index/Constraint Creation via schema query
+=== Script Option
 
-Following an example of how to create indexes while you're creating nodes
+The script option allow you to execute a series of preparation script before Spark
+Job execution, the result of the last query can be reused in combination with the
+`query` ingestion mode as it follows
 
 ----
+val ds = Seq(SimplePerson("Andrea", "Santurbano")).toDS()
+
 ds.write
-      .format(classOf[DataSource].getName)
-      .mode(SaveMode.Overwrite)
-      .option("url", SparkConnectorScalaSuiteIT.server.getBoltUrl)
-      .option("labels", ":Person:Customer")
-      .option("node.keys", "surname")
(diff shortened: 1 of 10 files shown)


src/main/scala/org/neo4j/spark/service/SchemaService.scala
```diff
@@ -331,18 +331,102 @@ class SchemaService(private val options: Neo4jOptions, private val driverCache:
...
+        case OptimizationType.CONSTRAINT => {
+          val assertType = if (props.size > 1) "NODE KEY" else "UNIQUE"
+          s"ON (${Neo4jUtil.NODE_ALIAS}:$quotedLabel) ASSERT ($quotedProps) IS $assertType"
+        }
+      }
+      val status = try {
+        session.run(s"$queryPrefix $querySuffix")
+        "CREATED"
+      } catch {
+        case _: Throwable => "KEPT"
```
moxious: not sure how this exception works but in 4.0 indexes fail to create if you do it twice, is this trying to catch the case where the index already existed?  In 3.5, index creation is idempotent

src/main/scala/org/neo4j/spark/service/SchemaService.scala
```diff
@@ -331,18 +331,102 @@ class SchemaService(private val options: Neo4jOptions, private val driverCache:
...
+    } else {
+      val quotedLabel = label.quote()
+      val quotedProps = props.map(prop => s"${Neo4jUtil.NODE_ALIAS}.${prop.quote()}").mkString(", ")
+      val actionName = s"spark_${action}_$label".quote()
+      val queryPrefix = s"CREATE ${action.toString} $actionName"
+      val querySuffix = action match {
+        case OptimizationType.INDEX => s"FOR (${Neo4jUtil.NODE_ALIAS}:$quotedLabel) ON ($quotedProps)"
+        case OptimizationType.CONSTRAINT => {
+          val assertType = if (props.size > 1) "NODE KEY" else "UNIQUE"
+          s"ON (${Neo4jUtil.NODE_ALIAS}:$quotedLabel) ASSERT ($quotedProps) IS $assertType"
```
moxious: this only supports node indexes but not relationship type property existence constraints -- also it doesn't support

ASSERT EXISTS (R.propertyName)

doc/docs/modules/ROOT/pages/quickstart_writer.adoc
```diff
@@ -215,4 +215,88 @@ UNwIND $events AS event
...
+----
+
+*So please into consideration that the first label is used for the index creation*
+
+===== Index/Constraint Creation via schema query
+
+Following an example of how to create indexes while you're creating nodes
+
+----
+ds.write
```
moxious, line 399: this is a bit different than the ticket asks for; here you do a bunch of queries ahead of doing a write, but it's all part of the write.   In this mode, it would be very hard to handle failure of idempotent index creation for example.

Why tie it to the write or read, rather than making the querys another write/read?

doc/docs/modules/ROOT/pages/quickstart_writer.adoc
```diff
@@ -215,4 +215,88 @@ UNwIND $events AS event
...
+      .format(classOf[DataSource].getName)
+      .mode(SaveMode.Overwrite)
+      .option("url", SparkConnectorScalaSuiteIT.server.getBoltUrl)
+      .option("labels", ":Person:Customer")
+      .option("node.keys", "surname")
+      .option("schema.optimization.type", "CONSTRAINT")
+      .save()
+----
+
+This will create, before the import starts, the following schema query:
```
moxious, line 382: I am not sure about this approach for a couple of reasons:

- If the number of nodes is large, creating the index before you begin a write guarantees that index population starts, not that the index is ready to go, unless you poll/monitor the index
- User doesn't practically have a way of controlling index naming
- Complex logic....if an index already exists from outside of spark, this should make 2 of them with double the overhead for the database.
- If spark creates indexes for your automatically, there are already a number of kinds of index syntax not yet supported, and in the future there will be more.  This is an additional point of coupling to the database language, and it can change.    For example, what should happen if someone does a fancy new 4.2 supported index type on a 3.5 database when the spark surface makes it look you can do this?

src/main/scala/org/neo4j/spark/service/SchemaService.scala
conker84: I can change it by adding an existence check on top, it will work on `3.5` and `4.x`

src/main/scala/org/neo4j/spark/service/SchemaService.scala
conker84: Ok il change enum from `CONSTRAINTS` to `NODE_CONSTRAINTS` so we can add more syntax sugar in the future ok?

doc/docs/modules/ROOT/pages/quickstart_writer.adoc
conker84: Actually the will be no failure in case the index is already present we log the creation operation as `KEPT`

doc/docs/modules/ROOT/pages/quickstart_writer.adoc
conker84: > I am not sure about this approach for a couple of reasons:
> 
> * If the number of nodes is large, creating the index before you begin a write guarantees that index population starts, not that the index is ready to go, unless you poll/monitor the index

We can add a future option that waits for index to or we can add it in this PR if you want

> * User doesn't practically have a way of controlling index naming

True, but we use a naming convention `spark_<index_or_constraint>_<label>` making it recognisable
 
> * Complex logic....if an index already exists from outside of spark, this should make 2 of them with double the overhead for the database.

Sorry I don't get this can you elaborate?

> * If spark creates indexes for your automatically, there are already a number of kinds of index syntax not yet supported, and in the future there will be more.  This is an additional point of coupling to the database language, and it can change.    For example, what should happen if someone does a fancy new 4.2 supported index type on a 3.5 database when the spark surface makes it look you can do this?

This is just syntax sugar which is meant to simplify the index/constraint creation fro  for the user but we also are future proof because there is always the `QUERY` option where the developer can specify a set of schema write queries like these:

```scala
ds.write
      .format(classOf[DataSource].getName)
      .mode(SaveMode.Overwrite)
      .option("labels", ":Person:Customer")
      .option("node.keys", "surname")
      .option("schema.optimization.type", "QUERY")
      .option("schema.optimization.query", """CREATE INDEX ON :Person(surname);
          |CREATE CONSTRAINT ON (p:Product)
          | ASSERT (p.name, p.sku)
          | IS NODE KEY
          |""".stripMargin)
      .save()
```

conker84:
@moxious can you please take a look at it?

src/main/scala/org/neo4j/spark/service/SchemaService.scala
```diff
@@ -331,20 +330,134 @@ class SchemaService(private val options: Neo4jOptions, private val driverCache:
...
+        }
+      }
+      val actionName = s"spark_${creationType}_$label".quote()
+      val queryPrefix = s"CREATE $creationType $actionName"
+
+      val queryCheck = """CALL db.indexes() YIELD labelsOrTypes, properties, uniqueness
+          |WHERE labelsOrTypes = $labels AND properties = $properties AND uniqueness = $uniqueness
+          |RETURN count(*) > 0 AS isPresent""".stripMargin
+
+      val isPresent = session.run(queryCheck, Map("labels" -> Seq(label).asJava,
```
moxious: this can fail, and in some strange circumstance it probably will.  Guard with try/catch log error and then don't create the index.

src/main/scala/org/neo4j/spark/service/SchemaService.scala
```diff
@@ -331,20 +330,134 @@ class SchemaService(private val options: Neo4jOptions, private val driverCache:
...
+        false
+      }
+    }
   }
 
+  private def createIndexOrConstraint(action: OptimizationType.Value, label: String, props: Seq[String]) = {
+    if (action == OptimizationType.NONE) {
+      log.info("No optimization type provided")
+      Unit
+    } else {
```
moxious: remove else block.   It makes code easier to return if up top you have the shortcut return nothing statement, then the rest of the method just flows without the else block.   Alternatively, you have a lot of other functions below that are just match statements on an input.  This one could be done the same, but the double nested try/catch/if/else blocks below make the code harder to read.

src/main/scala/org/neo4j/spark/service/SchemaService.scala
```diff
@@ -331,20 +330,134 @@ class SchemaService(private val options: Neo4jOptions, private val driverCache:
...
+          "properties" -> props.asJava,
+          "uniqueness" -> uniqueness).asJava)
+        .single()
+        .get("isPresent")
+        .asBoolean()
+
+      val status = if (isPresent) {
+        "KEPT"
+      } else {
+        val query = s"$queryPrefix $querySuffix"
```
moxious: this chunk would be better refactored into a reusable function

- Log a message
- Run a query
- Log a failure if it happens

(this gets done in several places)

doc/docs/modules/ROOT/pages/quickstart_writer.adoc
```diff
@@ -319,4 +319,107 @@ You can specify 3 different modes to use for saving the nodes:
 
 * `Overwrite`: will perform a `MERGE` on that node
 * `ErrorIfExists`: will perform a `CREATE`
-* `Match`: will perform a `MATCH`
\ No newline at end of file
+* `Match`: will perform a `MATCH`
+
```
moxious, line 323: Not the right place here, but somewhere in the docs you need to document the "script" option that you added to this PR since it was opened.

moxious (commented):
some smaller stuff and then it's mergeable.

src/main/scala/org/neo4j/spark/service/SchemaService.scala
conker84: done

src/main/scala/org/neo4j/spark/service/SchemaService.scala
conker84: done, but I did it in a "scala way" with pattern matching

src/main/scala/org/neo4j/spark/service/SchemaService.scala
conker84: is it ok if we do it in a separate PR?

doc/docs/modules/ROOT/pages/quickstart_writer.adoc
conker84: You can find the documentation here:
https://github.com/neo4j-contrib/neo4j-spark-connector/blob/3b687d6dd0f130d90b12e7fcf29ef5d200d5ffd1/doc/docs/modules/ROOT/pages/quickstart_writer.adoc#script-option

conker84:
@moxious thank you for the feedback, I updated the PR, please look at my comments!

src/main/scala/org/neo4j/spark/service/SchemaService.scala
moxious: yes that's fine

PR #190, opened 2020-09-21
Fixes #180: Inner CALL { block } breaks certain cypher patterns when reading via custom query

### LINKED ISSUE #180: Inner CALL { block } breaks certain cypher patterns when reading via custom query
See screenshot:

Screen Shot 2020-09-10 at 12 33 44 PM

utnaf:
Fixes #180 

Additional details in the issue.

Please take a look at the newly added test for a use case.

### DIFF
--- a/src/main/scala/org/neo4j/spark/service/SchemaService.scala
+++ b/src/main/scala/org/neo4j/spark/service/SchemaService.scala
@@ -22,7 +22,6 @@ import scala.collection.mutable.ArrayBuffer
 object PartitionSkipLimit {
   val EMPTY = PartitionSkipLimit(0, -1, -1)
 }
-
 case class PartitionSkipLimit(partitionNumber: Int, skip: Long, limit: Long)
 
 class SchemaService(private val options: Neo4jOptions, private val driverCache: DriverCache, private val filters: Array[Filter] = Array.empty)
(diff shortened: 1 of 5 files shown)


src/main/scala/org/neo4j/spark/service/SchemaService.scala
```diff
@@ -130,32 +130,27 @@ class SchemaService(private val options: Neo4jOptions, private val driverCache:
...
-                |MATCH (${Neo4jUtil.RELATIONSHIP_SOURCE_ALIAS})-[${Neo4jUtil.RELATIONSHIP_ALIAS}:${options.relationshipMetadata.relationshipType}]->(${Neo4jUtil.RELATIONSHIP_TARGET_ALIAS})
-                |RETURN ${Neo4jUtil.RELATIONSHIP_ALIAS}
-                |ORDER BY rand()
-                |LIMIT ${options.schemaMetadata.flattenLimit}
-                |""".stripMargin
-              val params = Collections.emptyMap[String, AnyRef]()
-              retrieveSchema(query, params, { record => record.get(Neo4jUtil.RELATIONSHIP_ALIAS).asRelationship.asMap.asScala.toMap })
-            }
-          }
+          // TODO get back to Cypher DSL when rand function will be available
```
conker84, line 131: Please add a log warn message that explains that we switch to the query resolution.
Do the same for nodes.

conker84 (changes requested):
@utnaf please look at the comment

src/main/scala/org/neo4j/spark/service/SchemaService.scala
```diff
@@ -41,6 +41,7 @@ class SchemaService(private val options: Neo4jOptions, private val driverCache:
         case e: ClientException =>
           e.code match {
             case "Neo.ClientError.Procedure.ProcedureNotFound" => {
```
conker84: can you remove also this as you did for relationships?

conker84: And also in any other situation that we use the error code check

src/main/scala/org/neo4j/spark/service/SchemaService.scala
```diff
@@ -41,6 +41,7 @@ class SchemaService(private val options: Neo4jOptions, private val driverCache:
...
+              log.warn("Switching to query schema resolution")
```
conker84:
```suggestion
              log.warn("Switching to query schema resolution because of the following exception:", e)
```

src/main/scala/org/neo4j/spark/service/SchemaService.scala
conker84: same as above

conker84 (changes requested):

conker84 (approved):

conker84:
@utnaf thank you

PR #196, opened 2020-09-27
fixes #172: Add support for PushDownRequiredColumn

### LINKED ISSUE #172: Add support for PushDownRequiredColumn
Add support for PushDownRequiredColumn. 

This can reduce the amount of data moved from Neo4j to Spark by reducing the number of columns selected by Neo4j.

utnaf:
Fixes #172 

This can reduce the amount of data moved from Neo4j to Spark by reducing the number of columns selected by Neo4j.

### DIFF
--- a/src/main/scala/org/neo4j/spark/reader/Neo4jDataSourceReader.scala
+++ b/src/main/scala/org/neo4j/spark/reader/Neo4jDataSourceReader.scala
@@ -2,18 +2,20 @@ package org.neo4j.spark.reader
 
 import java.util
 
+import org.apache.spark.internal.Logging
 import org.apache.spark.sql.catalyst.InternalRow
 import org.apache.spark.sql.sources.Filter
 import org.apache.spark.sql.sources.v2.DataSourceOptions
 import org.apache.spark.sql.sources.v2.reader.{DataSourceReader, InputPartition, SupportsPushDownFilters, SupportsPushDownRequiredColumns}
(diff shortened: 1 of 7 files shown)


utnaf:
@conker84 we need to discuss about this one: I've been struggling with a weird behaviour.
.
This scenario

```
      .option("relationship", "KNOWS")
      .option("relationship.source.labels", "Person")
      .option("relationship.target.labels", "Person")
      .load()

    df.filter("`<source>`.`id` = '14' AND `<target>`.`id` = '16'").count
```

Where we don't explicitly select any column, will cause the PushDownColumn interface to prune the following schema

```
StructType(StructField(<source>,MapType(StringType,StringType,true),false),
StructField(<target>,MapType(StringType,StringType,true),false))
```

The actual schema should be

```
StructType(StructField(<rel.id>,LongType,false), 
StructField(<rel.type>,StringType,false), 
StructField(<source>,MapType(StringType,StringType,true),false), 
StructField(<target>,MapType(StringType,StringType,true),false))
```

I don't understand why it does this. Seems like it considers any filtered column as selected. 
If I remove the `target` condition, the pruned column will just be `source`. 

It's an odd behaviour and needs to be documented if we want to proceed with this implementation.


In addition, I'm having errors on relationship reads with `nodes.map == true`.

conker84:
@utnaf & @moxious I'm closing this we'll back ASAP once we have more info about the behaviour

PR #177, opened 2020-09-09
Improve Docs

utnaf:
This PR is for creating the final doc pages.

http://5.9.211.195:8000/neo4j-spark-docs/1.0.0/

PR #197, opened 2020-09-28
Set relationship.nodes.map default to false

utnaf:
As stated, now the option `relationship.nodes.map` is default to `false`


utnaf:
Related doc commit https://github.com/neo4j-contrib/neo4j-spark-connector/pull/177/commits/5b0256904876a07cb23ed279b13c7384a588b183

conker84:
@utnaf Can you delete `.option("relationship.nodes.map", "false")`? Otherwise looks good thanks!

utnaf:
> @utnaf Can you delete `.option("relationship.nodes.map", "false")`? Otherwise looks good thanks!

@conker84 I would leave it explicit in the tests so if we change it in the feature we don't have to go through the tests to set the option again.

PR #208, opened 2020-09-30
Fix writing merge example query

darabos:
Thanks again, these docs are great. I just spotted this typo. Not sure why there's a diff on the last line. I used the GitHub web editor. I guess it wants to add a final newline.

### DIFF
--- a/doc/docs/modules/ROOT/pages/writing.adoc
+++ b/doc/docs/modules/ROOT/pages/writing.adoc
@@ -287,8 +287,8 @@ The generated query will be
 [source,cypher]
 ----
 UNWIND $events AS event
-MERGE (n:`Person`:`Customer` {name: event.name, surname: event.surname})
-SET n += event
+MERGE (n:`Person`:`Customer` {name: event.keys.name, surname: event.keys.surname})
+SET n += event.properties
 ----
 
 Here you must specify which columns of your Dataframe will be used as keys to match the nodes.
--- a/doc/docs/modules/ROOT/pages/writing.adoc
+++ b/doc/docs/modules/ROOT/pages/writing.adoc
@@ -288,7 +288,7 @@ The generated query will be
 ----
 UNWIND $events AS event
 MERGE (n:`Person`:`Customer` {name: event.name, surname: event.surname})
-SET n += event.properties
+SET n += event
 ----
 
 Here you must specify which columns of your Dataframe will be used as keys to match the nodes.
--- a/doc/docs/modules/ROOT/pages/writing.adoc
+++ b/doc/docs/modules/ROOT/pages/writing.adoc
@@ -287,7 +287,7 @@ The generated query will be
 [source,cypher]
 ----
 UNWIND $events AS event
-MERGE (n:`Person`:`Customer` {name: event.keys.name, surname: event.keys.surname})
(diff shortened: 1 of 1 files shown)


darabos:
Sorry, I've made a few more changes. I was trying to replicate this query for various reasons, and I know very little about Cypher. But it seems to work now.

doc/docs/modules/ROOT/pages/writing.adoc
```diff
@@ -286,9 +286,9 @@ The generated query will be
 
 [source,cypher]
 ----
-UNNIND $events AS event
-MERGE (n:`Person`:`Customer` {name: event.keys.name, surname: event.keys.surname})
-SET n += event.properties
+UNWIND $events AS event
+MERGE (n:`Person`:`Customer` {name: event.name, surname: event.surname})
```
conker84: why you removed the `keys` property?

conker84 (changes requested):

doc/docs/modules/ROOT/pages/writing.adoc
darabos: > why you removed the `keys` property?

Because there's no `keys` property. The DataFrame we are writing has Person objects. A Person has `name` and `surname`, but no `keys`.

(I mean I could be wrong. I'm using this API for the first time.)

utnaf: Hi @darabos , thanks for your PR. 

Regarding the `keys` property: I would say it's needed. The `$events` parameter is generated [here](https://github.com/neo4j-contrib/neo4j-spark-connector/blob/4.0/src/main/scala/org/neo4j/spark/service/MappingService.scala#L23-L42), in particular look at lines 29, where we set the `keys` property.

In addition, I just added a PR to the Spark Notebooks repo https://github.com/utnaf/spark-connector-notebooks/pull/4 with the merge example, so you can try it out if you want. 
If you enable Spark logs on Zeppelin logs you will see this:

```
Writing a batch of 3 elements to Neo4j,
for jobId=60b6dced-d8aa-4d0d-b823-cf6784675403 and partitionId=0
with query: UNWIND $events AS event
MERGE (node:Person:Musician {id: event.keys.id})
SET node += event.properties
```

I guess we can put the `keys` back =) Thank you!

utnaf (changes requested):

conker84:
@darabos PING

darabos:
> @darabos PING

Sorry! One, sec, I'll update the PR. Thanks for the comments.

doc/docs/modules/ROOT/pages/writing.adoc
darabos: > Regarding the `keys` property: I would say it's needed. The `$events` parameter is generated [here](https://github.com/neo4j-contrib/neo4j-spark-connector/blob/4.0/src/main/scala/org/neo4j/spark/service/MappingService.scala#L23-L42), in particular look at lines 29, where we set the `keys` property.

I see! So when we use a custom query, `keys` and `properties` are not there, but they are present when we let the connector generate the query.

> I guess we can put the `keys` back =) Thank you!

Done. I've put back `properties` too. Sorry for the delay.

moxious:
Thank you for submitting!

PR #220, opened 2020-10-08
fixes #209: Enable the script option for read access, not only write.

### LINKED ISSUE #209: Enable the "script" option for read access, not only write.
We have a "script" option, but it appears only accessible for writes, which require a dataframe.  I'd like them for reads too.  The key use case here is GDS -- where what the "script" would do is basically create a virtual graph and run an analysis, while the actual read query would return the analysis.   There's no writing that's going on with that, and if "script" is available write-only, then it means I'd have to invent a useless dataframe in order to use GDS.

Screen Shot 2020-10-01 at 6 56 11 AM

conker84:
Fixes #209 

The `script` option is now working also in `DataSourceReader`

### DIFF
--- a/src/test/scala/org/neo4j/spark/DataSourceReaderTSE.scala
+++ b/src/test/scala/org/neo4j/spark/DataSourceReaderTSE.scala
@@ -2,6 +2,7 @@ package org.neo4j.spark
 
 import java.sql.Timestamp
 import java.time.{LocalDateTime, OffsetDateTime, ZoneOffset}
+import java.util.Collections
 
 import org.apache.spark.sql.catalyst.expressions.GenericRowWithSchema
 import org.apache.spark.sql.{DataFrame, Row}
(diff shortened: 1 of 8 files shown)


src/test/scala/org/neo4j/spark/DataSourceReaderTSE.scala
```diff
@@ -1096,6 +1097,25 @@ class DataSourceReaderTSE extends SparkConnectorScalaBaseTSE {
     assertEquals(2, df.count())
   }
 
+  @Test
+  def testShouldPassTheScriptResult(): Unit = {
+    val df = ss.read
+      .format(classOf[DataSource].getName)
+      .option("url", SparkConnectorScalaSuiteIT.server.getBoltUrl)
+      .option("script", "RETURN 'foo' AS val")
```
moxious, line 1105: Nice.  Just a question on this point:  While partitions=2 here, does script get executed twice, once in each partition, or once upfront and then passed to each partition?  I think it gets executed twice, right?

moxious (changes requested):
Very good, close to merge.    Request that you document the script option in the docs.  It's already in on the write page, but not mentioned on the read page in the docs.

src/test/scala/org/neo4j/spark/DataSourceReaderTSE.scala
conker84: No, the query gets executed just once on the Spark Driver program an then the result is shared across partitions

moxious (approved):

PR #221, opened 2020-10-08
fixes #217: Multiple duplicate property names when reading multi-labeled nodes

### LINKED ISSUE #217: Multiple duplicate property names when reading multi-labeled nodes
See screenshot for code example.

Screen Shot 2020-10-06 at 10 05 09 AM

To reproduce, launch a Neo4j sandbox (GDS example) and just run the query in the screenshot.  The underlying issue is caused by multiply-labeled nodes.  I'm reading :Person nodes but there are :Person:Dead, :Person:Knight, and many other combinations.   This makes the results of apoc.meta.nodeTypeProperties() repeat the property over and over for each label combination, which I think is causing the issue.

This is just a placeholder issue.  The behavior is arguably correct right now, in that nodeLabels is coming out correct, and you have a different property per node label set, but it's difficult to use.  Need brainstorming on what better situation would be doable.

conker84:
fixes #217 

Multiple duplicate property names when reading multi-labelled nodes are now managed properly

### DIFF
--- a/src/main/scala/org/neo4j/spark/service/SchemaService.scala
+++ b/src/main/scala/org/neo4j/spark/service/SchemaService.scala
@@ -8,7 +8,7 @@ import org.apache.spark.sql.sources.Filter
 import org.apache.spark.sql.types.{DataType, DataTypes, StructField, StructType}
 import org.neo4j.driver.exceptions.ClientException
 import org.neo4j.driver.types.Entity
-import org.neo4j.driver.{Record, Session, Transaction, TransactionWork}
+import org.neo4j.driver.{Record, Session, Transaction, TransactionWork, Value}
 import org.neo4j.spark.service.SchemaService.{cypherToSparkType, normalizedClassName, normalizedClassNameFromGraphEntity}
 import org.neo4j.spark.util.Neo4jImplicits.{CypherImplicits, EntityImplicits}
 import org.neo4j.spark.util.{Neo4jUtil, ValidationUtil}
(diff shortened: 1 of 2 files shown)


src/main/scala/org/neo4j/spark/service/SchemaService.scala
```diff
@@ -33,7 +33,11 @@ class SchemaService(private val options: Neo4jOptions, private val driverCache:
 
   private def structForNode(labels: Seq[String] = options.nodeMetadata.labels): StructType = {
     var structFields: mutable.Buffer[StructField] = (try {
-        val query = "CALL apoc.meta.nodeTypeProperties({ includeLabels: $labels })"
+        val query =
+          """CALL apoc.meta.nodeTypeProperties({ includeLabels: $labels })
+            |YIELD propertyName, propertyTypes
+            |RETURN DISTINCT propertyName, propertyTypes
```
moxious: I think this won't work, because you'll miss type information.

Imagine this scenario:
* `(:Person { prop: 25 })`
* `(:Person:Player { prop: "hello" })`
* `(:Person:Player:Weirdo { prop: true })`

If you get the distinct propertyName, won't you get only one of those property type lines, basically at random?   If you want to take a distinct propertyName, you need to collect/merge all of the various types they can have.   In the good case, if you collect all of the various types you'll end up with [String, String, String] and then it's easy.   But you could still end up with [String, Long, Boolean]

moxious (changes requested):

src/main/scala/org/neo4j/spark/service/SchemaService.scala
moxious: If the connector gets wrong metadata it's going to be quite hard to debug, since APOC is only one of the techniques.  Then we will probably start to think that it's a problem with APOC, when it's about local user database state

conker84: That case we managed it case APOC is not present, we will coerce all to Strings, I think that we need to have the same strategy here, wdyt?

moxious: If the types differ, "coerce everything to string" is right.  But the types might be the same across the different labels.  So if the total type collection is [Long, Long, Long, Long] -- then of course use long.  ;)

conker84: sure I meant the use-case: `[String, Long, Boolean]` :D

conker84:
@moxious if there are no further comments I'll merge it

PR #229, opened 2020-10-16
Fixes #198: Improve query.count docs

### LINKED ISSUE #198: LIMIT doesn't work with custom read cypher query
See screenshot.

If you take out the LIMIT 10 this does work.  With the LIMIT 10 it's an error.

Screen Shot 2020-09-28 at 7 59 52 AM

utnaf:
Fixes #198

As discussed in the issue, we already have the option `query.count` we can use to limit the returned data by the cypher query.


utnaf:
Added control to check if SKIP/LIMIT are at the end of the query. Please see the test cases.

doc/docs/modules/ROOT/pages/reading.adoc
```diff
@@ -220,6 +220,32 @@ The created schema will be
 The returned column order is not guarantee to match the RETURN statement for Neo4j 3.* and Neo4j 4.0.
 Starting from Neo4j 4.1 it the order will be the same.
 
+[[bookmark-limit-query]]
+==== Limit the results
+
+Since partitioning is done by using LIMIT/SKIP you should not limit the result yourself.
```
moxious: Mini rewrite of this section, to include new information I just discovered.

This connector does not permit using SKIP or LIMIT at the end of a Cypher query.  Attempts to do this will result in errors, such as the message "Please provide a valid READ query".

This is not supported because internally, the connector uses SKIP/LIMIT pagination to break read sets up into multiple partitions, to support partitioned reads.  As a result, user-provided SKIP/LIMIT clashes with what the connector itself adds to your query to support parallelism.

There is a work-around though; you can still accomplish the same by using SKIP/LIMIT internal inside of the query, rather than after the final RETURN block of the query.

Here's a simple example.   This first query will be rejected, and will fail:

```
MATCH (p:Person)
RETURN p.name as name
ORDER BY name
LIMIT 10
```

This same query though can be reformulated and will still work.

```
MATCH (p:Person)
WITH p.name as name 
ORDER BY name 
LIMIT 10
RETURN p.name
```

The queries return the exact same data, but only the second one is usable with the spark connector, and partition-able, because of the `WITH` clause, and the simple final `RETURN` clause.   If you choose to reformulate queries to use "internal SKIP/LIMIT" take careful notice of ordering operations to guarantee the same result set.

You may also use the `query.count` option rather than reformulating your query (rest of your text here)

moxious (changes requested):

doc/docs/modules/ROOT/pages/reading.adoc
utnaf: > This connector does not permit using SKIP or LIMIT at the end of a Cypher query. Attempts to do this will result in errors, such as the message "Please provide a valid READ query".

fyi now it says: `SKIP/LIMIT are not allowed at the end of the query`

changed the rest.

src/main/scala/org/neo4j/spark/util/Validations.scala
```diff
@@ -110,6 +110,8 @@ object Validations {
             s"You need to set the ${Neo4jOptions.RELATIONSHIP_TARGET_LABELS} option")
         }
         case QueryType.QUERY => {
+          ValidationUtil.isFalse(neo4jOptions.query.value.matches("(?s).*(LIMIT \\d+|SKIP ?\\d+)\\s*\\z"),
```
conker84: i guess we need to do the matches as insensitive?

conker84: please also add a test with the skip/limit as snake case and lowercase

conker84 (changes requested):

conker84 (approved):

moxious (approved):

src/main/scala/org/neo4j/spark/util/Validations.scala
drumcircle: Note sure this is fixed.  It complains wherever I put "LIMIT" in this query and query.count does not limit results:

```
relationship_cypher="""
MATCH (s)-[r:IS_SIMILAR]-(t)
RETURN s.name AS source
,t.name AS target
,r.similarity AS similarity
,s.communityId AS louvain_community_id
"""
relsDf2=spark.read.format("org.neo4j.spark.DataSource") \
    .option("query", relationship_cypher) \
    .option("url", neo4j_url) \
    .option("authentication.type", "basic") \
    .option("authentication.basic.username", neo4j_user) \
    .option("authentication.basic.password", neo4j_password) \
    .option("query.count", 2) \
    .load()
    
relsDf2
```

Issue #199, opened 2020-09-28
Wrong error message when providing a database option on 3.5


moxious:
See example.  This works without the "database" option.  With, it fails with the incorrect error message.

Very important: this is a 3.5 cluster I'm attaching to.  So the use of the database option is *wrong*.   We should either ignore the option (if that's possible) or fail with a better message than this.

<img width="1058" alt="Screen Shot 2020-09-28 at 8 12 11 AM" src="https://user-images.githubusercontent.com/5471710/94430717-5b975800-0162-11eb-900d-14d5e6531c55.png">

moxious:
This may be fixed by surfacing driver / session errors.

This query can't succeed because it doesn't make sense (3.5 isn't multidatabase)

Check to see if session throws an error (I bet it does) -- if this error is properly surfaced to the user, then you're done.

utnaf:
I think #183 will fix this as well. 

But when we add the option consistency check this should be solved anyways.

moxious:
We think this is fixed due to previous PR but I need to verify to close.

moxious:
@utnaf this is not fixed by the recent PR merge.   Here's another replication case, now with a different error message.

The endpoint is 3.5.11 causal cluster with 3 core & 1 read-replica.  I can create a replication environment if you need.

```
spark.read.format("org.neo4j.spark.DataSource").option("url", "neo4j://REMOVED:7687")
    .option("authentication.type", "basic")
    .option("authentication.basic.username", "neo4j")
    .option("authentication.basic.password", "(REMOVED)")
    .option("database", "foo")
    .option("query", "MATCH (h:Household) RETURN id(h)").load().show()
```

This code provides a correct result *without* the database option, but with the database option, provides this error:

Result:

```
java.lang.IllegalArgumentException: Please provide a valid READ query
  at org.neo4j.spark.util.ValidationUtil$.isTrue(ValidationUtil.scala:22)
  at org.neo4j.spark.util.Validations$$anonfun$4.apply(Validations.scala:113)
  at org.neo4j.spark.util.Validations$$anonfun$4.apply(Validations.scala:93)
  at org.neo4j.spark.reader.Neo4jDataSourceReader$$anonfun$1.apply(Neo4jDataSourceReader.scala:22)
  at org.neo4j.spark.reader.Neo4jDataSourceReader$$anonfun$1.apply(Neo4jDataSourceReader.scala:22)
[...]

PR #231, opened 2020-10-20
Fixes #170: add connection check on session creation

### LINKED ISSUE #170: Add connection check before reading and writing
When writing ad dataset to neo4j, if there is a driver misconfiguration we have this error:

```
org.apache.spark.SparkException: Writing job aborted.
  at org.apache.spark.sql.execution.datasources.v2.WriteToDataSourceV2Exec.doExecute(WriteToDataSourceV2Exec.scala:92)
  at org.apache.spark.sql.execution.SparkPlan$$anonfun$execute$1.apply(SparkPlan.scala:131)
  at org.apache.spark.sql.execution.SparkPlan$$anonfun$execute$1.apply(SparkPlan.scala:127)
  at org.apache.spark.sql.execution.SparkPlan$$anonfun$executeQuery$1.apply(SparkPlan.scala:155)
  at org.apache.spark.rdd.RDDOperationScope$.withScope(RDDOperationScope.scala:151)
  at org.apache.spark.sql.execution.SparkPlan.executeQuery(SparkPlan.scala:152)
  at org.apache.spark.sql.execution.SparkPlan.execute(SparkPlan.scala:127)
  at org.apache.spark.sql.execution.QueryExecution.toRdd$lzycompute(QueryExecution.scala:80)
  at org.apache.spark.sql.execution.QueryExecution.toRdd(QueryExecution.scala:80)
  at org.apache.spark.sql.DataFrameWriter$$anonfun$runCommand$1.apply(DataFrameWriter.scala:676)
  at org.apache.spark.sql.DataFrameWriter$$anonfun$runCommand$1.apply(DataFrameWriter.scala:676)
  at org.apache.spark.sql.execution.SQLExecution$$anonfun$withNewExecutionId$1.apply(SQLExecution.scala:78)
  at org.apache.spark.sql.execution.SQLExecution$.withSQLConfPropagated(SQLExecution.scala:125)
  at org.apache.spark.sql.execution.SQLExecution$.withNewExecutionId(SQLExecution.scala:73)
  at org.apache.spark.sql.DataFrameWriter.runCommand(DataFrameWriter.scala:676)
  at org.apache.spark.sql.DataFrameWriter.save(DataFrameWriter.scala:260)
```

that doesn't say too much about why the writing failed. 
The actual error can be seen down the stack trace but I think it's easier to perform a connection test before trying to read or write data, this way we can show a better error message.

utnaf:
Fixes #170 and #199 

Added a connection check on session creation. Please tell me if there is a better way to check the connectivity.

This is the now error mentioned in #199 

Screenshot 2020-10-20 at 09 46 57

### DIFF
--- a/src/main/scala/org/neo4j/spark/util/Validations.scala
+++ b/src/main/scala/org/neo4j/spark/util/Validations.scala
@@ -2,7 +2,7 @@ package org.neo4j.spark.util
 
 import org.apache.spark.sql.SaveMode
 import org.apache.spark.sql.types.StructType
-import org.neo4j.driver.AccessMode
+import org.neo4j.driver.{AccessMode, Session}
 import org.neo4j.spark.service.{Neo4jQueryStrategy, SchemaService}
 import org.neo4j.spark.util.Neo4jImplicits.StructTypeImplicit
 import org.neo4j.spark._
@@ -15,6 +15,7 @@ object Validations {
     val cache = new DriverCache(neo4jOptions.connection, jobId)
     val schemaService = new SchemaService(neo4jOptions, cache)
     try {
+      validateConnection(cache.getOrCreate().session(neo4jOptions.session.toNeo4jSession))
       neo4jOptions.query.queryType match {
         case QueryType.QUERY => {
           ValidationUtil.isTrue(schemaService.isValidQuery(
@@ -94,6 +95,7 @@ object Validations {
     val cache = new DriverCache(neo4jOptions.connection, jobId)
     val schemaService = new SchemaService(neo4jOptions, cache)
     try {
+      validateConnection(cache.getOrCreate().session(neo4jOptions.session.toNeo4jSession))
       neo4jOptions.query.queryType match {
         case QueryType.LABELS => {
           ValidationUtil.isNotEmpty(neo4jOptions.nodeMetadata.labels,
@@ -129,6 +131,15 @@ object Validations {
     } finally {
       schemaService.close()
       cache.close()
+    }
+  }
+
+  def validateConnection(session: Session): Unit = {
+    try {
+      session.run("EXPLAIN RETURN 1").consume()
+    }
+    finally {
+      session.close()
     }
   }
 }
--- a/src/test/scala/org/neo4j/spark/DataSourceReaderNeo4j35xTSE.scala
+++ b/src/test/scala/org/neo4j/spark/DataSourceReaderNeo4j35xTSE.scala
@@ -14,21 +14,41 @@ object DataSourceReaderNeo4j35xTSE {
 
 class DataSourceReaderNeo4j35xTSE extends SparkConnectorScalaBaseTSE {
   @Test
-  def testShouldThrowClearErrorIfACanComputeTheSchema(): Unit = {
+  def testShouldThrowClearErrorIfADbIsSpecified(): Unit = {
     try {
       ss.read.format(classOf[DataSource].getName)
         .option("url", SparkConnectorScalaSuiteIT.server.getBoltUrl)
-        .option("query", "MATCH (i:DO_NOT_EXIST) RETURN i")
+        .option("database", "db1")
+        .option("labels", "MATCH (h:Household) RETURN id(h)")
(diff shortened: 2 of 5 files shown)


src/main/scala/org/neo4j/spark/util/Validations.scala
```diff
@@ -129,4 +131,13 @@ object Validations {
       cache.close()
     }
   }
+
+  def validateConnection(session: Session): Unit = {
+    try {
+      session.run("EXPLAIN RETURN 1").consume()
```
moxious, line 139: Yep, that'd work.  Honestly you don't even need the explain, you could just roundtrip "something".   `RETURN 1` by itself would work, because it proves bi-directional communication and that the session is sensible, but either way is really fine, since you're not doing anything with the explain anyway

src/main/scala/org/neo4j/spark/util/Validations.scala
```diff
@@ -15,6 +15,7 @@ object Validations {
     val cache = new DriverCache(neo4jOptions.connection, jobId)
     val schemaService = new SchemaService(neo4jOptions, cache)
     try {
+      validateConnection(cache.getOrCreate().session(neo4jOptions.session.toNeo4jSession))
```
moxious, line 18: watch out on semantics; this method is saying "get or create" - is session reuse possible here?  Because validateConnection closes the session.   Not sure if there's a code path to be concerned about there, but you have to throw away this session after validating the connection and you can't reuse it.

src/main/scala/org/neo4j/spark/util/Validations.scala
```diff
@@ -94,6 +95,7 @@ object Validations {
     val cache = new DriverCache(neo4jOptions.connection, jobId)
     val schemaService = new SchemaService(neo4jOptions, cache)
```
moxious, line 96: A read is always tied to a set of options.

You have validateConnection called twice, in reader and in writer.   Rather than remembering to call it each time, if a SchemaService is tied to a set of options, you could put this into the initialization of the schemaservice.  Before you even get to the query, the schemaservice could validate connection for you, and the code would maybe be more dry, you don't have to remember to validate each time.

src/test/scala/org/neo4j/spark/DataSourceReaderNeo4j35xTSE.scala
```diff
@@ -0,0 +1,37 @@
...
+import org.neo4j.driver.exceptions.ClientException
+
+object DataSourceReaderNeo4j35xTSE {
+  @BeforeClass
+  def checkNeo4jVersion() {
+    Assume.assumeTrue(TestUtil.neo4jVersion().startsWith("3.5"))
+  }
+}
+
+class DataSourceReaderNeo4j35xTSE extends SparkConnectorScalaBaseTSE {
```
moxious, line 15: Another unit test scenario here which is the same problem but a different source, please add this and see if it behaves:

User connects to Neo4j 4.0 with databases (neo4j, system) and issues a query with `.option("database", "noexist")`.   Actually it's more likely they'll have like a typo in the database or whatever.

In 3.5 there is no valid database option, but in 4.0 you can always mess it up.  ;)

src/main/scala/org/neo4j/spark/util/Validations.scala
utnaf: True what you said, also @conker84 made the same observation. But I would rather validate the connection in the Validator than validate in the SchemaService. Yep the code will be more dry, but if I need to find the connection validation, the validator would be the first file I would look into.

I don't really have a strong opinion on this. I could change it if the majority agrees it's better the other way, I was just explaining why it is how it is

src/main/scala/org/neo4j/spark/util/Validations.scala
utnaf: `getOrCreate` is referred to the DriverCache, so what we're getting here is a driver instance and it's reusable. We can clear it up by chancing the `cache` variabile to `driverCache`

moxious (approved):

PR #236, opened 2020-10-26
Issue #172: Implement SupportsPushDownRequiredColumns

### LINKED ISSUE #172: Add support for PushDownRequiredColumn

utnaf:
Fixes #172

This can reduce the amount of data moved from Neo4j to Spark by reducing the number of columns selected by Neo4j.

### DIFF
--- a/src/main/scala/org/neo4j/spark/reader/Neo4jDataSourceReader.scala
+++ b/src/main/scala/org/neo4j/spark/reader/Neo4jDataSourceReader.scala
@@ -75,5 +75,7 @@ class Neo4jDataSourceReader(private val options: DataSourceOptions, private val 
 
   override def pushedFilters(): Array[Filter] = filters
 
-  override def pruneColumns(requiredSchema: StructType): Unit = requiredColumns = requiredSchema
+  override def pruneColumns(requiredSchema: StructType): Unit = {
+    requiredColumns = requiredSchema
+  }
 }
--- a/src/main/scala/org/neo4j/spark/reader/Neo4jDataSourceReader.scala
+++ b/src/main/scala/org/neo4j/spark/reader/Neo4jDataSourceReader.scala
@@ -8,7 +8,7 @@ import org.apache.spark.sql.sources.v2.DataSourceOptions
 import org.apache.spark.sql.sources.v2.reader.{DataSourceReader, InputPartition, SupportsPushDownFilters, SupportsPushDownRequiredColumns}
 import org.apache.spark.sql.types.StructType
 import org.neo4j.spark.{DriverCache, Neo4jOptions}
-import org.neo4j.spark.service.{PartitionSkipLimit, SchemaService}
+import org.neo4j.spark.service.SchemaService
 import org.neo4j.spark.util.Validations
 
 import scala.collection.JavaConverters._
@@ -62,7 +62,7 @@ class Neo4jDataSourceReader(private val options: DataSourceOptions, private val 
     // we generate a partition for each element
     partitionSkipLimitList
       .map(partitionSkipLimit => new Neo4jInputPartitionReader(neo4jOptions, filters, schema, jobId,
-        partitionSkipLimit, scriptResult))
+        partitionSkipLimit, scriptResult, requiredColumns))
   }
 
   override def pushFilters(filtersArray: Array[Filter]): Array[Filter] = {
@@ -76,10 +76,10 @@ class Neo4jDataSourceReader(private val options: DataSourceOptions, private val 
   override def pushedFilters(): Array[Filter] = filters
 
   override def pruneColumns(requiredSchema: StructType): Unit = {
-    requiredColumns = if (neo4jOptions.pushdownColumnsEnabled) {
-      requiredSchema
-    } else {
+    requiredColumns = if (!neo4jOptions.pushdownColumnsEnabled || neo4jOptions.relationshipMetadata.nodeMap) {
       new StructType()
+    } else {
+      requiredSchema
     }
   }
 }
--- a/src/main/scala/org/neo4j/spark/reader/Neo4jDataSourceReader.scala
+++ b/src/main/scala/org/neo4j/spark/reader/Neo4jDataSourceReader.scala
@@ -76,6 +76,10 @@ class Neo4jDataSourceReader(private val options: DataSourceOptions, private val 
   override def pushedFilters(): Array[Filter] = filters
 
   override def pruneColumns(requiredSchema: StructType): Unit = {
-    requiredColumns = requiredSchema
+    requiredColumns = if (neo4jOptions.pushdownColumnsEnabled) {
+      requiredSchema
+    } else {
+      new StructType()
+    }
   }
 }
--- a/src/main/scala/org/neo4j/spark/reader/Neo4jDataSourceReader.scala
+++ b/src/main/scala/org/neo4j/spark/reader/Neo4jDataSourceReader.scala
@@ -14,9 +14,12 @@ import org.neo4j.spark.util.Validations
 import scala.collection.JavaConverters._
 
 class Neo4jDataSourceReader(private val options: DataSourceOptions, private val jobId: String) extends DataSourceReader
-  with SupportsPushDownFilters {
+  with SupportsPushDownFilters
+  with SupportsPushDownRequiredColumns {
 
   private var filters: Array[Filter] = Array[Filter]()
+
+  private var requiredColumns: StructType = new StructType()
 
   private val neo4jOptions: Neo4jOptions = new Neo4jOptions(options.asMap())
     .validate(options => Validations.read(options, jobId))
@@ -71,4 +74,6 @@ class Neo4jDataSourceReader(private val options: DataSourceOptions, private val 
   }
 
   override def pushedFilters(): Array[Filter] = filters
+
+  override def pruneColumns(requiredSchema: StructType): Unit = requiredColumns = requiredSchema
 }
--- a/src/main/scala/org/neo4j/spark/service/Neo4jQueryService.scala
+++ b/src/main/scala/org/neo4j/spark/service/Neo4jQueryService.scala
@@ -212,11 +212,7 @@ class Neo4jQueryReadStrategy(filters: Array[Filter] = Array.empty[Filter],
       matchQuery.returning(entity)
     }
     else {
-      matchQuery.returning(requiredColumns.map {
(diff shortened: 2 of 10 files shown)


utnaf:
@moxious added a branch on the notebook repo with this branch JAR https://github.com/utnaf/spark-connector-notebooks/tree/push-down-columns

src/main/scala/org/neo4j/spark/service/Neo4jQueryService.scala
```diff
@@ -124,8 +124,44 @@ class Neo4jQueryReadStrategy(filters: Array[Filter] = Array.empty[Filter],
...
+          sourceNode
+        }
+        else if (entityName.contains(Neo4jUtil.RELATIONSHIP_TARGET_ALIAS)) {
+          targetNode
+        }
+        else {
+          throw new IllegalArgumentException(s"`${column}` is not a valid column.`")
+        }
+
+        if (splatColumn.length.equals(1)) {
```
conker84:
```suggestion
        if (splatColumn.length == 1) {
```

src/main/scala/org/neo4j/spark/service/Neo4jQueryService.scala
```diff
@@ -171,11 +207,32 @@ class Neo4jQueryReadStrategy(filters: Array[Filter] = Array.empty[Filter],
...
+  }
+
   override def createStatementForNodes(options: Neo4jOptions): String = {
     val node = createNode(Neo4jUtil.NODE_ALIAS, options.nodeMetadata.labels)
     val matchQuery = filterNode(node)
-    val returning = matchQuery.returning(node)
-    renderer.render(buildStatement(returning))
+//    val returning = matchQuery.returning(node)
+//    renderer.render(buildStatement(returning))
+    renderer.render(buildStatement(returnRequiredColumns(node, matchQuery)))
```
conker84: please expand it, is deeply nested

src/main/scala/org/neo4j/spark/reader/Neo4jDataSourceReader.scala
```diff
@@ -71,4 +74,6 @@ class Neo4jDataSourceReader(private val options: DataSourceOptions, private val
   }
 
   override def pushedFilters(): Array[Filter] = filters
+
+  override def pruneColumns(requiredSchema: StructType): Unit = requiredColumns = requiredSchema
```
conker84: please add a configuration param that enable/disable the pushdown columns as we did for pushdown filters

src/main/scala/org/neo4j/spark/service/Neo4jQueryService.scala
```diff
@@ -124,8 +124,44 @@ class Neo4jQueryReadStrategy(filters: Array[Filter] = Array.empty[Filter],
...
+
+    val returnExpressions: Seq[Expression] = if (requiredColumns.isEmpty) {
+      Seq(aliasedRelationship, aliasedSourceNode, aliasedTargetNode)
+    }
+    else {
+      requiredColumns.map(column => {
+        val splatColumn = column.split('.')
+        val entityName = splatColumn.head
+
+        val entity = if (entityName.contains(Neo4jUtil.RELATIONSHIP_ALIAS)) {
```
conker84, line 141: why contains and not `==`?

conker84: also for the others

src/main/scala/org/neo4j/spark/service/Neo4jQueryService.scala
```diff
@@ -226,7 +283,7 @@ class Neo4jQueryReadStrategy(filters: Array[Filter] = Array.empty[Filter],
     if (labels.isEmpty) {
       Cypher.anyNode(name)
     } else {
-      Cypher.node(primaryLabel, otherLabels.asJava).named(name)
+      Cypher.node(primaryLabel, otherLabels.asJava).named(name.quote())
```
conker84: are you sure that is necessary?

src/main/scala/org/neo4j/spark/service/Neo4jQueryService.scala
```diff
@@ -124,8 +124,44 @@ class Neo4jQueryReadStrategy(filters: Array[Filter] = Array.empty[Filter],
...
+          entity match {
+            case n: Node => n.as(entityName.quote())
+            case r: Relationship => r.getRequiredSymbolicName
+          }
+        }
+        else {
+          getCorrectProperty(column, entity)
+        }
+      })
+    }
```
conker84: can you please move these lines into a method?

src/test/scala/org/neo4j/spark/DataSourceReaderTSE.scala
```diff
@@ -1296,6 +1298,213 @@ class DataSourceReaderTSE extends SparkConnectorScalaBaseTSE {
...
+      .option("url", SparkConnectorScalaSuiteIT.server.getBoltUrl)
+      .option("relationship", "BOUGHT")
+      .option("relationship.source.labels", "Product")
+      .option("relationship.target.labels", "Person")
+      .load
+      .select("`source.name`", "`<source.id>`")
+
+    df.count()
+
+    assertEquals(Set("source.name", "<source.id>"), df.columns.toSet)
```
conker84, line 1386: `Seq` instead of `Set`? we want also to check the order here

conker84: same as in the other tests

conker84: maybe here you can leave the tests as they already are and add similar in 4.x?

conker84 (changes requested):
Thank you @utnaf please look at my comments!

src/test/scala/org/neo4j/spark/DataSourceReaderTSE.scala
utnaf: problem is that in 3.5 order is not guaranteed, I will do the same test in 4.x

src/main/scala/org/neo4j/spark/service/Neo4jQueryService.scala
utnaf: because the name can be `<source>` or `source`. 

Example:
`source.name` or `<source.id>`

conker84: what happens there is a filed named `sourceOfTruth`?

utnaf: it can't be named like that in the relationship strategy... you just have `source.` `target.` `relationship.`, if you select any other field name it breaks before since it doesn't match the schema.

src/test/scala/org/neo4j/spark/service/Neo4jQueryServiceTest.scala
```diff
@@ -99,6 +159,69 @@ class Neo4jQueryServiceTest {
...
+    options.put(Neo4jOptions.URL, "bolt://localhost")
+    options.put("relationship", "KNOWS")
+    options.put("relationship.nodes.map", "false")
+    options.put("relationship.source.labels", "Person")
+    options.put("relationship.target.labels", "Person")
+    val neo4jOptions: Neo4jOptions = new Neo4jOptions(options)
+
+    val query: String = new Neo4jQueryService(neo4jOptions, new Neo4jQueryReadStrategy(
+      Array[Filter](),
+      PartitionSkipLimit(0, -1, -1),
```
conker84: please use `PartitionSkipLimit.EMPTY`, also in the other places

src/test/scala/org/neo4j/spark/service/Neo4jQueryServiceTest.scala
```diff
@@ -188,7 +334,7 @@ class Neo4jQueryServiceTest {
       "WHERE ((source.name = 'John Doe' OR target.name = 'John Doraemon' OR source.name = 'Jane Doe') " +
       "AND (target.age = 34 OR target.age = 18) " +
       "AND rel.score = 12) " +
-      "RETURN source, rel, target", query)
+      "RETURN rel, source AS source, target AS target", query)
```
conker84, line 337: why we have this?

src/test/scala/org/neo4j/spark/service/Neo4jQueryServiceTest.scala
```diff
@@ -138,7 +261,30 @@ class Neo4jQueryServiceTest {
...
+    val options: java.util.Map[String, String] = new java.util.HashMap[String, String]()
+    options.put(Neo4jOptions.URL, "bolt://localhost")
+    options.put("relationship", "KNOWS")
+    options.put("relationship.nodes.map", "true")
+    options.put("relationship.source.labels", "Person")
+    options.put("relationship.target.labels", "Person")
+    val neo4jOptions: Neo4jOptions = new Neo4jOptions(options)
+
+    val filters: Array[Filter] = Array[Filter](
+      EqualTo("source.id", "14"),
```
conker84, line 278: can you change it to an int instead of a string? just to check if we handle the where correctly

conker84 (changes requested):

src/test/scala/org/neo4j/spark/service/Neo4jQueryServiceTest.scala
utnaf: it would require additional logic to set or not set an alias if the names are equal. I can add it but I don't think it's worth it since this is invisible to the user and / or causes overhead on the query performance

src/test/scala/org/neo4j/spark/service/Neo4jQueryServiceTest.scala
utnaf: what do you mean? if we change to int here it will be an int in the query and result will be like 

`MATCH (source:`Person`) MATCH (target:`Person`) MATCH (source)-[rel:`KNOWS`]->(target) WHERE (source.id = 14 AND target.id = 16) RETURN rel, source AS source, target AS target`

utnaf: I can add a case in the IT for the DataSourceReader if there is none

conker84 (approved):

PR #241, opened 2020-11-02
fixes #175: Ensure that apoc.meta.(node|rel)TypeProperties parameters can be passed to control sampling

### LINKED ISSUE #175: Ensure that apoc.meta.(node|rel)TypeProperties parameters can be passed to control sampling
In this schema inference technique, user needs to be able to specify control properties.  The default sampling properties work well in most situations, but when the database is extremely large, they will need to be adjusted to avoid over-sampling.

conker84:
Fixes #175 

`apoc.meta.(node|rel)TypeProperties` configuration parameters can now be managed via `option` method.

- added feature
- added test
- added docs

### DIFF
--- a/doc/docs/modules/ROOT/pages/architecture.adoc
+++ b/doc/docs/modules/ROOT/pages/architecture.adoc
@@ -16,7 +16,86 @@ the two systems are based on completely different organizing data models:
 
 The use of this connector can be thought of as a "Transform" step and a "Load" step, regardless of
 which direction the data is moving.  The "Transform" step is concerned with how to move data between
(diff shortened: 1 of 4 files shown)


moxious (commented):
This looks good to me, but as usual, we need a bit more documentation.   👍 

We just need to describe why you'd want to do this, not just how to do it.   Maybe in the secton that describes how apoc.meta.nodeTypeProperties is used, add a note saying that these parameter tunings here are for adjusting performance of that on large databases, or something like this.

conker84:
> apoc.meta.nodeTypeProperties

Done!

moxious (approved):

Issue #243, opened 2020-11-05
[nodemon] clean exit - waiting for changes before restart


beliefer:
```
[gengjiaan@qsslog1 doc]$ npm run start

> @neo4j/docs@1.0.0 start /home/**/my-fork/neo4j-spark-connector/doc
> npm run server && npm-watch


> @neo4j/docs@1.0.0 server /home/**/my-fork/neo4j-spark-connector/doc
> forever start server.js

warn:    --minUptime not set. Defaulting to: 1000ms
warn:    --spinSleepTime not set. Your script will exit if it does not stay up for at least 1000ms
info:    Forever processing file: server.js
(node:31916) Warning: Accessing non-existent property 'padLevels' of module exports inside circular dependency
(Use `node --trace-warnings ...` to show where the warning was created)
(node:31916) Warning: Accessing non-existent property 'padLevels' of module exports inside circular dependency
No task specified. Will go through all possible tasks
[build:docs] [nodemon] 2.0.6
[build:docs] [nodemon] clean exit - waiting for changes before restart
```

jexp:
Do you have any network connection issues to github?
Perhaps a proxy setup or so for network connections?

beliefer:
> Do you have any network connection issues to github?
> Perhaps a proxy setup or so for network connections?

I resolved the network issue, but keep stuck in `[build:docs] [nodemon] clean exit - waiting for changes before restart`.

adam-cowley:
That is the output from the build process - if you make a change to an .adoc file that should re-run.  Do you also see a message saying that there is a server running on http://localhost:8000?

beliefer:
> That is the output from the build process - if you make a change to an .adoc file that should re-run. Do you also see a message saying that there is a server running on http://localhost:8000?

I visited http://localhost:8000 but got the message:
`Cannot GET /neo4j-spark-docs/1.0.0`

adam-cowley:
It looks like the redirect on the root is broken, try http://localhost:8000/spark instead.

beliefer:
> It looks like the redirect on the root is broken, try http://localhost:8000/spark instead.

Thanks! your suggestion fix my issue.

PR #247, opened 2020-11-09
Fix redirect for local preview

adam-cowley:
Fixes the local docs preview as mentioned in #243


conker84 (approved):
thank you @adam-cowley

Issue #206, opened 2020-09-30
Compatibility with Spark 3.0


usmanmunara:
Hi, 

I have been trying to connect Spark 3.0 with Neo4j 4.1. However, the connector doesn't seem to work, it's throwing quite a lot of errors. Before sharing the specific errors I have I was just curious if neo4j-spark-connector is ported to support Spark 3.0 with Neo4j 4.1. 

Thanks

moxious:
Related issue: https://github.com/neo4j-contrib/neo4j-spark-connector/issues/203

We're in the process of updating the connector to work with the Spark 2.4 dataframes API.   3.0 is still not supported because it is relatively new and not widely adopted, and had breaking backwards compatibility changes.   Please see the latest release -- we just put one out yesterday, that has the latest available code.

wangying0420:
Our cluster uses spark3.0. I believe many companies have upgraded to the latest spark. I think it is necessary to be compatible with spark3.0

bbenzikry:
To add to the discussion on adoption, spark 3 support is the only reason we're not using this great project and rely on convoluted methods to load data from our tables due to upstream dependencies that simply don't work well with spark 2.

2.x support in frameworks is heavily dwindling. An example from databricks is delta lake
![image](https://user-images.githubusercontent.com/1993348/96948300-f8dd7600-14ed-11eb-967b-1ab62233bc64.png)

Really looking forward for a spark 3 release
P.S great work 👍 I really like the direction this is taking and just had to say I really appreciate the effort and hard work you're all investing in this

moxious:
Need to consider release strategy going forward when this work is done.   JARs which are released need to be versioned according to:

- Connector version (4.0.0)
- Scala version (2.1.1)
- Spark version (2.4 vs. 3.0)

mvn sub-modules; business logic vs. spark specifics.  Spark 3.0 deprecated scala 2.11 & added support for scala 2.13

utnaf:
Hi @usmanmunara, we just merged the Spark 3.0 support! We are not releasing it yet since we need to do some additional testing, but please feel free to use the attached jar to play around and let us know if you have any issue!

Be aware that `SaveMode.ErrorIfExists` is not supported at the moment for Spark 3.0. 

[...]

PR #282, opened 2021-01-25
#206: Split current code into common and spark-2.4 modules

utnaf:
In order to be able to add Spark 3.0 support (#206) we are first opening a PR where we introduce maven modules project structure. Once this is merged we'll add the Spark 3.0 module. 

Current problems:

- [x] <del>When packaged, a `doc` directory is created inside each module (maven frontend plugin's fault)</del>


moxious:
needs review -- does this keep antora build pipeline in good shape?

moxious:
@conker84 to review and merge

conker84 (approved):
@utnaf generally looks good, IDK if makes sense implement my suggestion

conker84:
ok for some reason seems that GitHub ignored my comment, I'll post it here:

IDK if makes sense to create a `test-support` module like we did in Streams:
https://github.com/neo4j-contrib/neo4j-streams/tree/4.0/test-support
In order to reuse test classes like `TestUtils` and `Neo4jContainerExtension` on all modules

conker84:
Thinking about it, it makes totally sense there are duplications like `Neo4jContainerExtension` that we want to avoit, please add the `test-support module`

conker84 (changes requested):
se the comment above

utnaf:
@conker84 updated

spark-2.4/pom.xml
```diff
@@ -0,0 +1,38 @@
...
+        <dependency>
+            <groupId>neo4j-contrib</groupId>
+            <artifactId>neo4j-connector-apache-spark_${scala.binary.version}_common</artifactId>
+            <version>4.0.0</version>
+        </dependency>
+
+        <dependency>
+            <groupId>neo4j-contrib</groupId>
+            <artifactId>neo4j-connector-apache-spark_${scala.binary.version}_test-support</artifactId>
+            <version>4.0.0</version>
```
conker84, line 34:
```suggestion
            <version>4.0.0</version>
            <scope>test</test>
```

conker84 (changes requested):
Just that change we need to import it as test package otherwise we'll include them into the distribution artifact.

conker84:
LGTM

I'm merging this so we can move forward to `3.0` integration, thank you @utnaf

conker84 (approved):

PR #290, opened 2021-02-15
Fixes #266: added test to ensure arrives to the Neo4j Driver

### LINKED ISSUE #266: Verify LDAP Basic Authentication
See https://neo4j.com/docs/driver-manual/current/client-applications/

Authentication type basic can work with LDAP, but need to verify how the token is formed and how that is expressed in terms of login.

utnaf:
Fixes #266 

Tried to check that the correct token is injected into the `ConfigurationSettings` [class](https://github.com/neo4j/neo4j-java-driver/blob/4.3/driver/src/main/java/org/neo4j/driver/internal/DriverFactory.java#L106) but wasn't able to.

Here we ensure that the driver is created properly by the `GraphDatabase` class.


common/src/test/java/org.neo4j.spark.service/AuthenticationTest.java
```diff
@@ -0,0 +1,60 @@
...
+import static org.junit.Assert.assertEquals;
+import static org.mockito.ArgumentMatchers.*;
+import static org.mockito.Mockito.times;
+
+@PrepareForTest({GraphDatabase.class})
+@RunWith(PowerMockRunner.class)
+public class AuthenticationTest {
+
+    @Test
+    public void testLdapConnectionToken() {
```
conker84: @utnaf can we write this in scala?

utnaf: Nope, can get it working in Scala, sadly there is an issue with power mock that doesn't make it work

utnaf: Jk, it was the other test I tried to do that wasn't working in scala, this one is. Will move it to the scala package.

test-support/pom.xml
```diff
@@ -49,6 +53,18 @@
             <version>2.12.0</version>
         </dependency>
 
+        <dependency>
+            <groupId>org.powermock</groupId>
+            <artifactId>powermock-module-junit4</artifactId>
+            <version>${powermock.version}</version>
```
conker84, line 59: add both as test deps

test-support/pom.xml
```diff
@@ -49,6 +53,18 @@
...
+        </dependency>
+
+        <dependency>
+            <groupId>org.powermock</groupId>
+            <artifactId>powermock-api-mockito2</artifactId>
+            <version>${powermock.version}</version>
```
conker84, line 65: test dep

conker84 (changes requested):

conker84:
LGTM thank you @utnaf

PR #296, opened 2021-02-26
Fixes #213: Add Python tests

### LINKED ISSUE #213: Add Python Tests
Add Python Tests

utnaf:
Todo

- [x] write tests
- [x] automate tests
- [x] add to CI


pom.xml
```diff
@@ -58,6 +58,7 @@
         <cypherdsl.version>2020.1.4</cypherdsl.version>
         <node.version>v12.16.0</node.version>
         <npm.version>6.13.7</npm.version>
+        <java.version>11</java.version>
```
conker84: We cannot do that because we still need to support Java 8.
Please look at:
https://spark.apache.org/docs/3.0.0/

_Spark runs on Java 8/11_

.github/workflows/python-spark2.4-tests.yml
```diff
@@ -0,0 +1,41 @@
...
+      - 'master'
+  pull_request:
+    branches:
+      - '*'
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.6, 3.7]
```
conker84: it would be great add tests over different neo4j version, like here:
https://github.com/neo4j-contrib/neo4j-spark-connector/blob/422bf30486fb0c40f638631c12b89cdc17eadb57/.github/workflows/ci.yml#L16

scripts/python/test_spark.py
```diff
@@ -0,0 +1,306 @@
...
+
+if len(sys.argv) != 3:
+    print("Not enough arguments")
+    sys.exit()
+
+scala_version = str(sys.argv.pop())
+spark_version = str(sys.argv.pop())
+
+if __name__ == "__main__":
+    with Neo4jContainer('neo4j:4.2') as neo4j_container:
```
conker84: take the neo4j version from the arguments as well

spark-2.4/pom.xml
```diff
@@ -33,6 +33,14 @@
             <version>4.0.0</version>
             <scope>test</scope>
         </dependency>
+
+        <!-- this is the only way i found to make stuff work in Intellij -->
+        <dependency>
```
conker84: I think we don't need this anymore

.github/workflows/python-spark3.0-tests.yml
```diff
@@ -0,0 +1,34 @@
+name: Python Tests for Spark 3.0
```
conker84: I think that we can merge the two actions into one just by adding the spark version into the matrix, please look at:
https://github.com/neo4j-contrib/neo4j-spark-connector/pull/301/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR15

conker84: and then you can manage what you need to do inside the python script

conker84 (changes requested):
@utnaf really impressing stuff thank you so much for doing this, I think that will boost a lot the coverage.
Please look at my comments.

.github/workflows/python-tests.yml
```diff
@@ -0,0 +1,43 @@
...
+        spark-version: [ {short: 2.4, ext: 2.4.5 }, {short: 3.0, ext: 3.0.1} ]
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Set up JDK 8
+        if: ${{ matrix.spark-version.short == 2.4}}
```
conker84: why this?

.github/workflows/python-tests.yml
```diff
@@ -0,0 +1,43 @@
...
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install pypandoc
+          pip install pyspark==${{ matrix.spark-version.ext }} "testcontainers[neo4j]"
+      - name: Build artifact
+        run: |
+          ./mvnw clean package -DskipTests -P spark-${{ matrix.spark-version }} -P scala-2.12 --no-transfer-progress
+      - name: Run tests for Spark ${{ matrix.spark-version.ext }} and Neo4j ${{ matrix.neo4j-version }}
+        if: ${{ ! (matrix.spark-version.short == 2.4 && matrix.python-version == 3.8) }}
```
conker84, line 39: @utnaf maybe makes sense add it also in other steps?

conker84 (commented):
@utnaf please look at my questions

.github/workflows/python-tests.yml
utnaf: we use java 11 for spark 3.0

.github/workflows/python-tests.yml
utnaf: mmm... idk, I'd rather find a cleaner solution...like having 2 jobs, one for spark 3 and one for spark 2.4

conker84: I don't like having a situation like that except for the case we have different python scripts for `3.0` and `2.4`, but seems not the case, the problem here is that we cannot leverage the `if:` with `matrix` directly into the `build` branch, there is a feature request into the Gh Actions repo, I hope that they will fix it asap

conker84:
Thank you so much @utnaf !!!!

PR #301, opened 2021-03-02
fixes #260: Move CI system to GitHub Actions

### LINKED ISSUE #260: Move CI system to GitHub Actions
Investigate

conker84:
- [x] Moved the CI to github actions
- [x] improved test code by refactoring and put in common some shared parts


.github/workflows/jvm-ci.yml
```diff
@@ -0,0 +1,38 @@
...
+    branches: [ '**' ]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        scala-version: [2.11, 2.12]
+        spark-version: ["2.4", "3.0"]
```
utnaf, line 16: here the version are as string but later (lines 21 and 23) you check them as float, is it correct? can we have them both as float?

utnaf (changes requested):
Just one small thing, plus can we delete the `debug.log` file or add it to the gitignore?

Amazing job though!

.github/workflows/jvm-ci.yml
conker84: yes it's the only way that works as expected because if I use `spark-version: [2.4, 3.0]` in case of `3.0` the value got coerced into `3` 🤷‍♂️

conker84:
thanks for catching the `debug.log`, I deleted it

.github/workflows/jvm-ci.yml
utnaf: ok, I'm pretty sure it works for me on the python tests

PR #306, opened 2021-03-09
Fixes #284: leverage Spark Config as default configuration values

### LINKED ISSUE #284: Leverage Spark Config as default configuration values
In order to avoid to the user to re-insert every time the common configurations such as connection parameters we can leverage the `SparkConf` via the current spark session in this way:

```scala
val defaultConfigMap = SparkSession.getActiveSession
        .map { _.sparkContext.getConf
          .getAllWithPrefix("neo4j.")
          .map { elem => (elem._1.substring("neo4.".length), elem._2) }
          .toMap
        }
        .getOrElse(Map.empty)
```

This is the mapping between `DataSourceOptions` and `SparkConf`:
| DataSourceOptions | SparkConf |
| ------------- | ------------- |
| url | neo4j.url  |
| authentication.basic.username  | neo4j.authentication.basic.username  |
| authentication.basic.password  | neo4j.authentication.basic.password  |
| database  | neo4j.database  |

...and so on...

The values from `SparkConf` became the default values for the `DataSourceOptions` in case they are not defined.

This will simplify the management of recurring configurations such as database url, user and password.

utnaf:
Fixes #284

Had to change the code a bit since the configurations set via `ss.conf.set()` are not available in the sparkContext.

### DIFF
--- a/common/src/main/scala/org/neo4j/spark/util/Neo4jOptions.scala
+++ b/common/src/main/scala/org/neo4j/spark/util/Neo4jOptions.scala
@@ -1,19 +1,37 @@
 package org.neo4j.spark.util
 
-import org.apache.spark.sql.SaveMode
+import org.apache.spark.sql.{SaveMode, SparkSession}
 import org.neo4j.driver.Config.TrustStrategy
 import org.neo4j.driver._
 
 import java.io.File
 import java.net.URI
+import java.util
 import java.util.UUID
 import java.util.concurrent.TimeUnit
 import scala.collection.JavaConverters._
 
 
-class Neo4jOptions(private val parameters: java.util.Map[String, String]) extends Serializable {
+class Neo4jOptions(private val options: java.util.Map[String, String]) extends Serializable {
   import Neo4jOptions._
   import QueryType._
+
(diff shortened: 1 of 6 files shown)


common/src/main/scala/org/neo4j/spark/util/Neo4jOptions.scala
```diff
@@ -1,20 +1,38 @@
...
 import java.util.concurrent.TimeUnit
 import scala.collection.JavaConverters._
 
 
-class Neo4jOptions(private val parameters: java.util.Map[String, String]) extends Serializable {
+class Neo4jOptions(private val options: java.util.Map[String, String]) extends Serializable {
   import Neo4jOptions._
   import QueryType._
 
+  private def parameters: util.Map[String, String] = {
```
utnaf, line 19: could't find a better way to merge two immutable Maps

conker84: now it's better!

conker84:
@utnaf thank you!

conker84 (approved):

PR #308, opened 2021-03-11
Fixes #285: Add version check for Spark

### LINKED ISSUE #285: Leverage Spark Version in order to return a clear error in case the is an unsupported version
We can access to the Spark version via:

```
SparkSession.getActiveSession
        .map { _.version }
        .getOrElse("UNKNOWN")
```

utnaf:
Fixes #285 

The check between Spark 3 and Spark 2 (and viceversa) doesn't work properly. 
The version check is the entry point of the connector and is executed when the class DataSource is instantiated. Problem is the Spark 3 doesn't have some classes required by DataSource on Spark 2, and viceversa. In this was the DataSource is never instantiated and the process fails with an error like:

```
Caused by: ClassNotFoundException: org.apache.spark.sql.sources.v2.ReadSupport
```

The check is working between Spark 2.4 and Spark 2.3.

I left the check on connector for Spark 3 as well, throwing an exception if the Spark version is different from 3.0 or 3.1.

I added a FAQ in the docs explaining that type of error means there's a version conflict between Spark and the connector.

### DIFF
--- a/common/src/main/scala/org/neo4j/spark/util/Validations.scala
+++ b/common/src/main/scala/org/neo4j/spark/util/Validations.scala
@@ -9,14 +9,14 @@ import org.neo4j.spark.util.Neo4jImplicits.StructTypeImplicit
 
 object Validations extends Logging {
 
-  def version(supportedVersions: Seq[String]): Unit = {
+  def version(supportedVersions: String*): Unit = {
     val sparkVersion = SparkSession.getActiveSession
       .map { _.version }
       .getOrElse("UNKNOWN")
 
     ValidationUtil.isTrue(
-      sparkVersion == "UNKNOWN" || supportedVersions.contains(sparkVersion.split('.').slice(0, 2).mkString(".")),
-      s"""You current Spark version ${sparkVersion} is not supported by the current connector.
+      sparkVersion == "UNKNOWN" || supportedVersions.exists(sparkVersion.matches),
+      s"""Your currentSpark version ${sparkVersion} is not supported by the current connector.
        |Please visit https://neo4j.com/developer/spark/overview/#_spark_compatibility to know which connector version you need.
        |""".stripMargin
     )
--- a/common/src/main/scala/org/neo4j/spark/util/Validations.scala
+++ b/common/src/main/scala/org/neo4j/spark/util/Validations.scala
@@ -15,7 +15,7 @@ object Validations extends Logging {
       .getOrElse("UNKNOWN")
 
     ValidationUtil.isTrue(
-      sparkVersion != "UNKNOWN" && supportedVersions.contains(sparkVersion.split('.').slice(0, 2).mkString(".")),
+      sparkVersion == "UNKNOWN" || supportedVersions.contains(sparkVersion.split('.').slice(0, 2).mkString(".")),
       s"""You current Spark version ${sparkVersion} is not supported by the current connector.
        |Please visit https://neo4j.com/developer/spark/overview/#_spark_compatibility to know which connector version you need.
        |""".stripMargin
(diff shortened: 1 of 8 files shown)


common/src/main/scala/org/neo4j/spark/util/Validations.scala
```diff
@@ -1,17 +1,27 @@
...
 object Validations extends Logging {
 
+  def version(supportedVersions: Seq[String]): Unit = {
+    val sparkVersion = SparkSession.getActiveSession
+      .map { _.version }
+      .getOrElse("UNKNOWN")
+
+    ValidationUtil.isTrue(
+      sparkVersion == "UNKNOWN" || supportedVersions.contains(sparkVersion.split('.').slice(0, 2).mkString(".")),
+      s"""You current Spark version ${sparkVersion} is not supported by the current connector.
```
conker84: `Your current...`?

spark-3/src/main/scala/org/neo4j/spark/DataSource.scala
```diff
@@ -8,11 +8,13 @@ import org.apache.spark.sql.types.StructType
 import org.apache.spark.sql.util.CaseInsensitiveStringMap
 import org.neo4j.spark.service.SchemaService
 import org.neo4j.spark.util.Validations.validateConnection
-import org.neo4j.spark.util.{DriverCache, Neo4jOptions}
+import org.neo4j.spark.util.{DriverCache, Neo4jOptions, Validations}
 
 class DataSource extends TableProvider
   with DataSourceRegister {
 
+  Validations.version(Seq("3.0", "3.1"))
```
conker84: Why simply just `3`?
```suggestion
  Validations.version(Seq("3.1"))
```

common/src/main/scala/org/neo4j/spark/util/Validations.scala
```diff
@@ -1,17 +1,27 @@
...
 import org.neo4j.driver.{AccessMode, Session}
 import org.neo4j.spark.service.{Neo4jQueryStrategy, SchemaService}
 import org.neo4j.spark.util.Neo4jImplicits.StructTypeImplicit
-import org.neo4j.spark._
-
-import java.util.concurrent.Callable
 
 object Validations extends Logging {
 
+  def version(supportedVersions: Seq[String]): Unit = {
```
conker84: Using varargs make the api more clearer:
```suggestion
  def version(supportedVersions: String*): Unit = {
```

common/src/main/scala/org/neo4j/spark/util/Validations.scala
```diff
@@ -1,17 +1,27 @@
...
+    val sparkVersion = SparkSession.getActiveSession
+      .map { _.version }
+      .getOrElse("UNKNOWN")
+
+    ValidationUtil.isTrue(
+      sparkVersion == "UNKNOWN" || supportedVersions.contains(sparkVersion.split('.').slice(0, 2).mkString(".")),
```
conker84: I would change it in this way
```suggestion
      sparkVersion == "UNKNOWN" || supportedVersions.find(supportedVersion => version.matches(supportedVersion))
      .map(s => true)
      .getOrElse(false),
```

conker84 (changes requested):
thank you so much @utnaf can you please take a look to my comments?

common/src/main/scala/org/neo4j/spark/util/Validations.scala
utnaf: `supportedVersions.exists(sparkVersion.matches)` looks better?

conker84 (approved):
LGTM

Issue #262, opened 2020-12-02
Add support for Spark Structured Streaming API


utnaf:
It might be useful to add support for Structured Streaming API for both Spark 2.4 and Spark 3.0 (#206)

https://spark.apache.org/docs/latest/structured-streaming-programming-guide.html

https://spark.apache.org/docs/2.4.5/structured-streaming-programming-guide.html

fyi @conker84

moxious:
Write to Neo4j -> direction seems fairly straightforward; the table is unbounded in size, so you keep spooling transactions at Neo4j and it effectively doesn't end until the stream drains.  That seems fine, but raises questions about the monitoring of the stream throughput over time, something to think about.

Read from <- Neo4j is different; this is possibly more connected to reactive streams coming from Neo4j.  There might be missing neo4j features to make this really good; for example very cool would be "read from trigger".  Neo4j pushes records out on every commit (like kafka) and the driver has a reactive stream to read from.   This direction is interesting but seems much harder.

Some open questions:  would this be needed to write streams from external sources to Neo4j?  What options would users have to coalesce streams into windows and write those as regular dataframes?   Why or why not is that prohibitively difficult?   I.e. imagine we didn't do the feature.  My guess is that users would time window the stream, and then just deal with those as regular dataframes.  How much work is that?  When would it just plain not work (while still staying in scale bounds of what Neo4j can do?)

moxious:
Sink part -- POC in progress

PR #322, opened 2021-04-08
#262 part 1/4: Spark 24 Sink

utnaf:
This PR contains the Sink implementation for the Spark 2.4 Structured Streaming API.


common/src/main/scala/org/neo4j/spark/writer/BaseDataWriter.scala
```diff
@@ -91,15 +93,25 @@ abstract class BaseDataWriter(jobId: String,
...
 
+  /**
+   * df: we check if the thrown exception is STOPPED_THREAD_EXCEPTION. This is the
+   * exception that is thrown when the streaming query is interrupted, we don't want to cause
+   * any error in this case. The transaction are rolled back automatically.
+   */
   private def logAndThrowException(e: Exception): Unit = {
-    if (e.isInstanceOf[ClientException]) {
-      log.error(s"Cannot commit the transaction because: ${e.getMessage}")
+    if(e.isInstanceOf[ServiceUnavailableException] && e.getMessage == STOPPED_THREAD_EXCEPTION_MESSAGE) {
```
conker84, line 109: In this catch maybe makes sense check what kind of root cause has the exception, if is `org.apache.spark.TaskKilledException` we can log the message otherwise we throw the exception

common/src/main/scala/org/neo4j/spark/writer/BaseDataWriter.scala
```diff
@@ -91,15 +93,25 @@ abstract class BaseDataWriter(jobId: String,
...
+      logInfo(e.getMessage)
     }
     else {
-      log.error("Cannot commit the transaction because the following exception", e)
-    }
+      if (e.isInstanceOf[ClientException]) {
```
conker84, line 113: maybe makes sense eventually change the if in this way:
```suggestion
      if ((e.isInstanceOf[ServiceUnavailableException] && e.getMessage == STOPPED_THREAD_EXCEPTION_MESSAGE) || e.isInstanceOf[ClientException]) {
```

and remove the the if above if we cannot leverage the exception root cause

conker84 (changes requested):
@utnaf thank you so much! just two small consideration that are worth to investigate

common/src/main/scala/org/neo4j/spark/writer/BaseDataWriter.scala
```diff
@@ -91,15 +94,27 @@ abstract class BaseDataWriter(jobId: String,
...
-      log.error(s"Cannot commit the transaction because: ${e.getMessage}")
+    if(e.isInstanceOf[ServiceUnavailableException] && e.getMessage == STOPPED_THREAD_EXCEPTION_MESSAGE) {
+      // df: leaving it here so we make sure if the cause is an instance of TaskKilledException
+      logInfo(s"+++ Cause: ${e.getCause.toString}")
+      logInfo(e.getMessage)
     }
     else {
-      log.error("Cannot commit the transaction because the following exception", e)
-    }
+      if ((e.isInstanceOf[ServiceUnavailableException] && e.getMessage == STOPPED_THREAD_EXCEPTION_MESSAGE) || e.isInstanceOf[ClientException]) {
```
conker84: (e.isInstanceOf[ServiceUnavailableException] && e.getMessage == STOPPED_THREAD_EXCEPTION_MESSAGE)
this is always false in this section right?

common/src/main/scala/org/neo4j/spark/writer/BaseDataWriter.scala
```diff
@@ -91,15 +94,27 @@ abstract class BaseDataWriter(jobId: String,
...
+   * df: we check if the thrown exception is STOPPED_THREAD_EXCEPTION. This is the
+   * exception that is thrown when the streaming query is interrupted, we don't want to cause
+   * any error in this case. The transaction are rolled back automatically.
+   */
   private def logAndThrowException(e: Exception): Unit = {
-    if (e.isInstanceOf[ClientException]) {
-      log.error(s"Cannot commit the transaction because: ${e.getMessage}")
+    if(e.isInstanceOf[ServiceUnavailableException] && e.getMessage == STOPPED_THREAD_EXCEPTION_MESSAGE) {
+      // df: leaving it here so we make sure if the cause is an instance of TaskKilledException
+      logInfo(s"+++ Cause: ${e.getCause.toString}")
```
conker84: please remove this

common/src/main/scala/org/neo4j/spark/writer/BaseDataWriter.scala
```diff
@@ -91,15 +94,27 @@ abstract class BaseDataWriter(jobId: String,
...
+      logInfo(e.getMessage)
```
conker84: I would change it to warn

conker84 (changes requested):

common/src/main/scala/org/neo4j/spark/writer/BaseDataWriter.scala
utnaf: yep

conker84 (dismissed):

conker84:
Thank you so much @utnaf

conker84:
I'm merging this as the problem is only for docs, thank you @utnaf

PR #330, opened 2021-04-22
fixes #323: Node keys are ignored when the field is a map

### LINKED ISSUE #323: Node keys are ignored when the field is a map
Given this DataFrame 
```
root
 |-- <rel.id>: long (nullable = false)
 |-- <rel.type>: string (nullable = false)
 |-- <source>: map (nullable = false)
 |    |-- key: string
 |    |-- value: string (valueContainsNull = true)
 |-- <target>: map (nullable = false)
 |    |-- key: string
 |    |-- value: string (valueContainsNull = true)
 |-- rel.roles: array (nullable = true)
 |    |-- element: string (containsNull = true)
```


```
+--------+----------+-------------------------------------------------------------------------------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+-----------------------------------+
|<rel.id>|<rel.type>|<source>                                                                       |<target>                                                                                                                                                                                                                                                                    |rel.roles                          |
+--------+----------+-------------------------------------------------------------------------------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+-----------------------------------+
|7       |ACTED_IN  |[name -> "Emil Eifrem", born -> 1978, <labels> -> ["Person"], <id> -> 8]       |[title -> "The Matrix", released -> 1999, <labels> -> ["Movie"], <id> -> 0, tagline -> "Welcome to the Real World"]                                                                                                                                                         |[Emil]                             |
|3       |ACTED_IN  |[name -> "Hugo Weaving", born -> 1960, <labels> -> ["Person"], <id> -> 4]      |[title -> "The Matrix", released -> 1999, <labels> -> ["Movie"], <id> -> 0, tagline -> "Welcome to the Real World"]                                                                                                                                                         |[Agent Smith]                      |
```

When I try to write the dataframe:

```
import org.apache.spark.sql.{SaveMode, SparkSession}


df2.write
  .mode(SaveMode.ErrorIfExists)
  .option("url", "bolt://neo4j:7687")
  .option("database","prova3")
   .option("authentication.type", "basic")
    .option("authentication.basic.username", "neo4j")
    .option("authentication.basic.password", "zeppelin")
  .option("relationship", "ACTED_IN")
    .option("relationship.source.labels", ":Person")
    .option("relationship.source.save.mode", "Overwrite")
    .option("relationship.source.node.keys", "`<source>`.`name`:name")
    .option("relationship.target.labels", ":Movie")
    .option("relationship.target.save.mode", "Overwrite")
    .option("relationship.target.node.keys", "`<target>`.`title`:title")
    .save()
```

I get the following error:

```
Caused by: org.apache.spark.SparkException: Job aborted due to stage failure: Task 0 in stage 91.0 failed 1 times, most recent failure: Lost task 0.0 in stage 91.0 (TID 91, localhost, executor driver): java.lang.IllegalArgumentException: Write failed due to the following errors:
 - Schema is missing `<target>`.`title` from option `relationship.target.node.keys`
 - Schema is missing `<source>`.`name` from option `relationship.source.node.keys`

The option key and value might be inverted.
```

conker84:
Fixes #323 for map data types you can now use the inner field as in `node.keys` or `node.properties`

### DIFF
--- a/common/src/test/scala/org/neo4j/spark/util/Neo4jImplicitsTest.scala
+++ b/common/src/test/scala/org/neo4j/spark/util/Neo4jImplicitsTest.scala
@@ -94,9 +94,11 @@ class Neo4jImplicitsTest {
 
   @Test
   def `struct should return true if contains fields`: Unit = {
-    val struct = StructType(Seq(StructField("is_hero", DataTypes.BooleanType), StructField("name", DataTypes.StringType)))
+    val struct = StructType(Seq(StructField("is_hero", DataTypes.BooleanType),
+      StructField("name", DataTypes.StringType),
+      StructField("fi``(╯°□°)╯︵ ┻━┻eld", DataTypes.StringType)))
 
-    assertEquals(0, struct.getMissingFields(Set("is_hero", "name")).size)
+    assertEquals(0, struct.getMissingFields(Set("is_hero", "name", "fi``(╯°□°)╯︵ ┻━┻eld")).size)
   }
 
   @Test
@@ -104,5 +106,20 @@ class Neo4jImplicitsTest {
     val struct = StructType(Seq(StructField("is_hero", DataTypes.BooleanType), StructField("name", DataTypes.StringType)))
 
     assertEquals(Set[String]("hero_name"), struct.getMissingFields(Set("is_hero", "hero_name")))
+  }
+
+  @Test
+  def `getMissingFields should handle maps`: Unit = {
+    val struct = StructType(Seq(
+      StructField("im", DataTypes.StringType),
+      StructField("im.a", DataTypes.createMapType(DataTypes.StringType, DataTypes.StringType)),
+      StructField("im.also.a", DataTypes.createMapType(DataTypes.StringType, DataTypes.StringType)),
+      StructField("im.not.a.map", DataTypes.StringType),
+      StructField("fi``(╯°□°)╯︵ ┻━┻eld", DataTypes.StringType)
+    ))
+
+    val result = struct.getMissingFields(Set("im.aMap", "`im.also.a`.field", "`im.a`.map", "`im.not.a.map`", "fi``(╯°□°)╯︵ ┻━┻eld"))
+
+    assertEquals(Set("im.aMap"), result)
   }
 }
--- a/common/src/main/scala/org/neo4j/spark/reader/BasePartitionReader.scala
+++ b/common/src/main/scala/org/neo4j/spark/reader/BasePartitionReader.scala
@@ -24,7 +24,7 @@ abstract class BasePartitionReader(private val options: Neo4jOptions,
   private val driverCache: DriverCache = new DriverCache(options.connection,
     if (partitionSkipLimit.partitionNumber > 0) s"$jobId-${partitionSkipLimit.partitionNumber}" else jobId)
 
-  private val query: String = new Neo4jQueryService(options, new Neo4jQueryReadStrategy(filters, partitionSkipLimit, requiredColumns.getFieldsName))
+  private val query: String = new Neo4jQueryService(options, new Neo4jQueryReadStrategy(filters, partitionSkipLimit, requiredColumns.fieldNames))
     .createQuery()
 
   private val mappingService = new MappingService(new Neo4jReadMappingStrategy(options, requiredColumns), options)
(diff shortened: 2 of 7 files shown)


common/src/test/scala/org/neo4j/spark/util/Neo4jImplicitsTest.scala
```diff
@@ -105,4 +107,19 @@ class Neo4jImplicitsTest {
...
+      StructField("im", DataTypes.StringType),
+      StructField("im.a", DataTypes.createMapType(DataTypes.StringType, DataTypes.StringType)),
+      StructField("im.also.a", DataTypes.createMapType(DataTypes.StringType, DataTypes.StringType)),
+      StructField("im.not.a.map", DataTypes.StringType),
+      StructField("fi``(╯°□°)╯︵ ┻━┻eld", DataTypes.StringType)
+    ))
+
+    val result = struct.getMissingFields(Set("im.aMap", "`im.also.a`.field", "`im.a`.map", "`im.not.a.map`", "fi``(╯°□°)╯︵ ┻━┻eld"))
+
+    assertEquals(1, result.size)
```
utnaf: should we check which one is the missing field? just to avoid false positives in the future

utnaf:
```scala
assertEquals("im.aMap", result(0))
```

I guess something like this

utnaf (changes requested):

common/src/test/scala/org/neo4j/spark/util/Neo4jImplicitsTest.scala
conker84: Yes it's better let me do it

conker84: Done

utnaf (approved):
This one was super tricky, great job @conker84!

conker84:
@Mogztter @adam-cowley hi! can you help me to understand why the docs are failing?
Thank you so much!!!

ggrossetie:
@conker84 `/graphgists` is missing in this context which is expected but we need to explicitly add `load graphgists` in the `skipPatterns`:

https://github.com/neo4j-contrib/neo4j-spark-connector/blob/1f44ad06af138f60d76130ec3f0d7478524b4d3e/doc/tasks/lint-links.js#L11-L35

conker84:
> @conker84 `/graphgists` is missing in this context which is expected but we need to explicitly add `load graphgists` in the `skipPatterns`:
> 
> https://github.com/neo4j-contrib/neo4j-spark-connector/blob/1f44ad06af138f60d76130ec3f0d7478524b4d3e/doc/tasks/lint-links.js#L11-L35

Thank you @Mogztter , is this something that we need to do in other projects too? Because in my knowledge we didn't change anything in the docs

ggrossetie:
I think you are seeing this issue because you are using the latest (unversioned) version of the UI:

https://github.com/neo4j-contrib/neo4j-spark-connector/blob/1f44ad06af138f60d76130ec3f0d7478524b4d3e/doc/docs.yml#L15

I've added a new entry in the menu (`/graphgists/`) for the upcoming GraphGists site in the UI (currently on master branch).

I would recommend to use a released version of the UI. You will need to update the version from time to time to get the latest features but on the bright side your build will be more stable.

Having said that, you're right when we will upgrade to the latest UI version we will need to update the lint task (check links) on other projects too.

Does it make sense?

conker84:
@Mogztter sorry I'm a little bit confused, we need to add the `skipPatterns` entry or we need to update the UI to the latest "stable" version?

ggrossetie:
> sorry I'm a little bit confused, we need to add the skipPatterns entry or we need to update the UI to the latest "stable" version?

Ultimately, you will need to do both 😉 

But right now you can fix this issue by using `https://s3-eu-west-1.amazonaws.com/static-content.neo4j.com/build/ui-bundle-v0.11.2.zip` instead of https://github.com/neo4j-contrib/neo4j-spark-connector/blob/1f44ad06af138f60d76130ec3f0d7478524b4d3e/doc/docs.yml#L15 

Since version 0.11.2 does not include graphgists, the check link error should disappear.

Alternatively, you can use `https://s3-eu-west-1.amazonaws.com/static-content.neo4j.com/build/ui-bundle-v0.12.0.zip` and add `load graphgists` in the `skipPattern`.

ggrossetie:
@conker84 https://github.com/conker84/neo4j-spark-connector/pull/2

conker84:
I'm merging this since is only a problem with docs

conker84:
I'm merging this since is only a problem with docs

PR #336, opened 2021-04-26
fixes #333: Remove Transaction code LockClientStopped

### LINKED ISSUE #333: Remove Transaction code LockClientStopped
Remove `Neo.TransientError.Transaction.LockClientStopped` so the users can leverage the inner retry mechanism when they get a transaction lock error

conker84:
- [ ] Removed blocking transient error codes `Neo.TransientError.Transaction.Terminated` and `Neo.TransientError.Transaction.LockClientStopped`
- [ ] Added `transaction.retry.timeout` which is the time in ms that the Connector should wait before retry

### DIFF
--- a/doc/docs/modules/ROOT/pages/writing.adoc
+++ b/doc/docs/modules/ROOT/pages/writing.adoc
@@ -104,6 +104,11 @@ The DataSource Writer has several options in order to connect and persist data i
 |3
 |No
 
+|`transaction.retry.timeout`
+|The time in ms that the Connector should wait before retry
+|0
+|No
+
 4+|*Node Specific Options*
 
 |`node.keys`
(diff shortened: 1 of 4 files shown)


doc/docs/modules/ROOT/pages/writing.adoc
```diff
@@ -104,6 +104,11 @@ The DataSource Writer has several options in order to connect and persist data i
 |3
 |No
 
+|`transaction.retry.timeout`
+|The time in ms that the Connector should wait before retry
```
moxious, line 108: rest of code OK but please be a bit more specific here about which issue we're addressing.  This will confuse people into thinking that the connector retries queries; it doesn't really.  Best here would be to cite the driver manual where it talks about TX retries, and explain this in terms of what the neo4j official driver is doing.

moxious (changes requested):

doc/docs/modules/ROOT/pages/writing.adoc
conker84: This is not a driver feature, I thought it would be useful add a delay in the case of locks, I also asked to @davidfauth just to understand if he could have benefited from this option and he agreed. If you want we can get rid of this, otherwise I can extend the description in this way:

_In case of transaction retry this is the time in ms that the Connector should wait before trying to perform the same write operation again_

PR #338, opened 2021-05-04
Fixes #337: Implemented user defined schema on Spark 2.4 and Spark 3

### LINKED ISSUE #337: Possible multi-partitioning happening when partitions(1) for read queries
Reported by a customer but also replicated by me.

```
# Create a GDS in-memory graph for authors.

cypher = """
  CALL gds.graph.create.cypher(
    'authors', 
    'MATCH (a:Author) RETURN id(a) as id',
    'MATCH (a1:Author)-[r1:AUTHOR]->(p:Paper)<-[r2:AUTHOR]-(a2:Author)
     WHERE 
       id(a1) < id(a2) AND 
       r1.index <= 4 AND 
       r2.index <=4 
     RETURN id(a1) as source, id(a2) as target, count(p) as weight
    ',
    {}
  ) YIELD graphName RETURN graphName
"""

spark.read.format("org.neo4j.spark.DataSource") \
  .format("org.neo4j.spark.DataSource") \
  .option("url", dbutils.widgets.get("url")) \
  .option("authentication.basic.username", dbutils.widgets.get("user")) \
  .option("authentication.basic.password", dbutils.widgets.get("password")) \
  .option("query", cypher) \
  .option("partitions", "1") \
  .load().show()
```

The error is this: 

```
org.apache.spark.SparkException: Job aborted due to stage failure: Task 0 in stage 30.0 failed 4 times, most recent failure: Lost task 0.3 in stage 30.0 (TID 166) (10.57.233.165 executor 0): org.neo4j.driver.exceptions.ClientException: Failed to invoke procedure `gds.graph.create.cypher`: Caused by: java.lang.IllegalArgumentException: A graph with name &#39;authors&#39; already exists.
```

Now, when you try to create duplicate graphs in GDS this error is predictable.   Note the use of `.options("partitions", "1")`.

We need to double check on this.  The error itself is OK and doesn't cause harm because the query does create the GDS virtual graph.  But this error suggests that something is being executed in parallel when we tell the connector to use a single partition.

Tested using Databricks runtime 8.2 which includes Spark 3.1.1 -- it's possible this may be related?

utnaf:
Fixes #337 

Implemented user defined schema on Spark 2.4 and Spark 3

```scala
      .schema(StructType(Seq(StructField("age", DataTypes.StringType)))) // <--
      .option("query", "MATCH (n:Person) RETURN n.age AS age")
      .load()
```

This allows to skip the schema generation, avoiding the run of an additional query.

### DIFF
--- a/doc/docs/modules/ROOT/pages/faq.adoc
+++ b/doc/docs/modules/ROOT/pages/faq.adoc
@@ -67,7 +67,7 @@ You can solve it by adding APOC to your Neo4j installation; this will remove the
 all the values for that field will be casted to String. This because Spark is not schema free,
 and need each column to always have the same type.
 
-You can read more <<quickstart.adoc#bookmark-read-known-problem, here>>.
+You can read more <<quickstart.adoc#read-known-problem, here>>.
 
 == The returned columns are not in the same order as I specified in the query
 
--- a/doc/docs/modules/ROOT/pages/faq.adoc
+++ b/doc/docs/modules/ROOT/pages/faq.adoc
@@ -95,3 +95,11 @@ Caused by: ClassNotFoundException: org.apache.spark.sql.sources.v2.ReadSupport
 
 This means that your Spark version doesn't match the Spark version on the connector.
 Please refer to xref:overview.adoc#_spark_compatibility[this page] to know which version you need.
+
+[[graph-already-exists]]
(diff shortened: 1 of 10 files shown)


conker84 (changes requested):
LGTM please add a paragraph to the documentation that explains how to use the feature

conker84 (approved):

conker84:
Thank you @utnaf !!!

PR #346, opened 2021-05-12
Fixes #343: add Streaming docs

### LINKED ISSUE #343: Add documentation for Spark Structured Streaming
As title says, add the documentation for using the connector with Spark Structured Streaming

View #262

utnaf:
Fixes #343 

Added Sink streaming doc. Missing source docs since it's not mergeable yet.


utnaf:
The build is failing because of a missing page that is not yet there since is being added in this PR.

doc/docs/modules/ROOT/pages/streaming.adoc
```diff
@@ -0,0 +1,42 @@
...
+  .option("save.mode", "ErrorIfExists")
+  .option("checkpointLocation", "/tmp/checkpoint/myCheckPoint")
+  .option("labels", "Person")
+  .option("node.keys", "value")
+  .start()
+----
+
+As said, you can use any writing strategy: link:writing.adoc#write-node[Labels], link:writing.adoc#write-rel[Relationship], or link:writing.adoc#write-query[Query].
+
+The only difference is that you must set the `checkpointLocation` and `save.mode` options.
```
moxious, line 89: what are checkpoints?  (Link docs if that's easier)

What format do they have?

What options do I have to work with checkpoints, like, can I put them on cloud storage or is this always a local file?

What happens if I have bad data quality in my stream, and a batch blows up?  Does the next continue, or does the streaming ingest fail?  Is this controllable?

utnaf: Totally fair questions, will update the docs today. I'll add a streaming example to the notebooks as well

utnaf: Updated and added a simple streaming sink example https://github.com/utnaf/neo4j-connector-apache-spark-notebooks/commit/d9f89e8b120375197006b3f3cc818e089ea03d99 to the notebooks

doc/docs/modules/ROOT/pages/streaming.adoc
```diff
@@ -0,0 +1,139 @@
...
+
+=== Streaming Property Name
+
+For the streaming to work, we need each record to have a property of type `timestamp`
+to leverage when reading new data from Neo4j to be sent to the stream.
+
+Behind the scenes the connector is building a query with a where clause that checks for the records that have
+`[timestampProperty] >= currentTimestamp()`.
+
+So it's required that each node that has the timestamp property.
```
moxious: I think you might note here that a neo4j data type (timestamp) is required here too or possibly date.  Dates and timestamps can be compared; but string-formatted dates "2001-09-11" will not work because of cypher type comparison

doc/docs/modules/ROOT/pages/streaming.adoc
```diff
@@ -0,0 +1,139 @@
...
+|Yes^*^
+
+4+|*Source*
+
+|`streaming.property.name`
+|The timestamp property used for batch reading. Read more xref:#_streaming_property_name[here].
+|_(none)_
+|Yes^*^
+
+|`streaming.from`
```
moxious, line 32: in this section you have thee short description in the table and then link to the full description of the property.  Do that here too.   Also, clarify the one line description, I found it confusing.   Maybe something like "This option is used to tell the connector whether to stream changes from the entire history of the database, or only from the present moment in time forward"

doc/docs/modules/ROOT/pages/streaming.adoc
```diff
@@ -0,0 +1,139 @@
...
+|This option is used to tell the connector from where to send data to the stream.
+
+**NOW**: start streaming from the moment the stream starts
+
+**ALL**: send all the data in the DB to the stream before reading new data
+|`ALL`
```
moxious: we should consider changing the default to be NOW.  If a person has a large db and is trying streaming for the first time, then when they repeatedly run their startup examples they will get a nasty large surprise I think.   Because people have a "first try" experience it might make sense to be lightweight by default.

moxious: Also the ALL behavior is very confusing in the way it operates it seems.   It does what it advertises, but every window of the function it seems to re-pull the entire database.  I would expect it pulls the entire database but from that point forward it only updates what's new, not re-sample the entire database every window?

This sample database only has a few records in it created with `unwind range(0, 2) as idx with idx create (:Test { idx:idx, lastUpdated: timestamp() });` once.

Screen Shot 2021-09-14 at 7 22 17 AM

moxious: notice how the `idx` values repeat

moxious: NOW has the same behavior, from the current point in time forward.   I started with those 4 sample records, and then just added some at a much higher IDX value.  You can see it got them going forward in time, but repeats them over and over.

Screen Shot 2021-09-14 at 7 24 23 AM

doc/docs/modules/ROOT/pages/streaming.adoc
```diff
@@ -0,0 +1,139 @@
...
+    .option("streaming.from", "NOW")
+    .load()
+
+// Memory streaming format writes the streamed data to a SparkSQL table
+query = stream.writeStream
+    .format("memory")
+    .queryName("testReadStream")
+    .start()
+
+spark
```
moxious: add to comment that this has to be in a separate code block, or executed seconds later.  If you execute so quickly in a sequential manner, it will come back with nothing because the stream above is not initialized yet.  This bit me and created a startup experience that the stream doesn't return anything. Probably my mistake for being a noob to streaming but it'll happen.

doc/docs/modules/ROOT/pages/streaming.adoc
```diff
@@ -0,0 +1,139 @@
...
+  .show()
+----
+
+=== Streaming Property Name
+
+For the streaming to work, we need each record to have a property of type `timestamp`
+to leverage when reading new data from Neo4j to be sent to the stream.
+
+Behind the scenes the connector is building a query with a where clause that checks for the records that have
+`[timestampProperty] >= currentTimestamp()`.
```
moxious: we need a tiny change here also WHERE timeStampProperty IS NOT NULL

I'm currently getting results back when I have null timestamps.  This is a problem because they're not directly comparable, which means they'll come back as a result of any query, and will always sort to the top.  I would argue we need to exclude anything where this property is missing, and then possibly introduce an option (later) if users want it.

doc/docs/modules/ROOT/pages/streaming.adoc
```diff
@@ -0,0 +1,139 @@
...
+
+So it's required that each node that has the timestamp property.
+The property name can be anything, just remember to set the `streaming.property.name` accordingly.
+
+=== Streaming From
+
+You can decide to stream all the data in the db, or just the new ones.
+To achieve this you can set the `streaming.from` option to one of these two values:
+
+* `ALL`: will read all the data that matches they given options, once read those, will read the ones that has `[timestampProperty] >= currentTimestamp()`. This is the **default value for this option**.
```
moxious: Your criteria here is exactly the same as NOW so there's a typo here somewhere.   You see it says it will read ones with this criteria, same as NOW, so this will confuse people, what's the difference?

Also make sure to add

AND [timeStampProperty] IS NOT NULL add here

doc/docs/modules/ROOT/pages/streaming.adoc
```diff
@@ -0,0 +1,139 @@
...
+* `NOW`: will start reading the data that has the `[timestampProperty] >= currentTimestamp()`
+
+=== Reading mode
```
moxious, line 166: put this at the top when you introduce streaming.  This confused me out of the gate, but it makes total sense.  I'm just a lazy user who tried to do stuff before actually reading the document.  Also as a lazy user I like the copypasta examples of how to do source/sink as early as possible, with the reference documentation describing why it works coming second.

moxious (commented):
will come back and do more

moxious:
What does this mean?  This doesn't do anything sensible, but wondering what we should say about this in the docs, if anything.

Notice the query doesn't include lastUpdated.  How do queries intersect with streaming.property.name?  I'm not sure....

Screen Shot 2021-09-14 at 2 03 20 PM

moxious:
type casting error when reading from a stream with a simple query.  Check the schema of the resulting dataframe, it's right....not sure what's happening here.

a variant of the error with casting to string occurs with any query that I come up with that returns a raw type.  It might be timestamp related?  IDK

**UPDATE** see next comment too it's the same issue.   You can repro this whenever you use a "streaming.property.name" field that is a datetime() and not a timestamp().  It seems to only like timestamps().

Check very carefully in the dataframe schema of this screenshot, and note that "time" is of type "timestamp".  This fails every time.   When I use a timestamp() in Neo4j, this comes through to the spark layer as a long and then it works.

Screen Shot 2021-09-14 at 2 08 26 PM

moxious:
Reading with relationships seems also not to work related to a type conversion error.  See screenshot.

Setup to populate this is:

```
unwind range(0, 4) as idx
create (:Thingy)-[:RELATED_TO { idx: idx, lastUpdated: datetime() }]->(:Thingy)
```

Check it out though:  if I change lastUpdated to `timestamp()` instead of datetime, it works!

As a user they're going to expect both orderable date/time fields to work

Screen Shot 2021-09-14 at 3 14 55 PM

moxious:
last comment and I'm done testing.   On the NOW mode of label, I can reproduce it repeating the same records over and over, but I got it so that it only produced records from the now point forward.   In both ALL and NOW, it seems to repeat records over and over though, so that's a serious thing to look into.

The timestamp coercion thing described above is the other major thing that I found.

Other than that.....

* We rely on partitioning and batch sizes a lot.  Can we add something to the streaming section that just notes whether or not these things even matter, or if they just don't apply at all?   I can specify a batch size, but IDK if it does anything at all.
* code definitely does work and I was able to make all 3 modes (label, rel, and query) work

doc/docs/modules/ROOT/pages/streaming.adoc
utnaf: Can you give me the code that causes the error? I really can't replicate the issue. Also I find odd that if the clause is `WHERE timestamp > 21321323` Neo4j will return also results with timestamp = null. Or am I missing something?

doc/docs/modules/ROOT/pages/streaming.adoc
utnaf: I can't really replicate this.

Screenshot 2021-09-15 at 12 21 40

Here you see the idx repeats but the <id> and lastUpdated are different.

Everything seems to work smoothly to me. I used your same configurations.

```python
graph_stream = (
    spark.readStream
    .option("url", "bolt://neo4j:7687")
    .option("authentication.type", "basic")
    .option("authentication.basic.username", "neo4j")
    .option("authentication.basic.password", "password")
    .option("streaming.property.name", "lastUpdated")
    .option("streaming.from", "ALL")
    .option("batch.size", 10)
    .option("labels", "Test")
    .load()
)
```

```
query = (graph_stream
  .writeStream
  .format("memory")
  .queryName("stream_output")
  .start())
```

utnaf:
> What does this mean? This doesn't do anything sensible, but wondering what we should say about this in the docs, if anything.
> 
> Notice the query doesn't include lastUpdated. How do queries intersect with streaming.property.name? I'm not sure....
> 
> Screen Shot 2021-09-14 at 2 03 20 PM

This might happen because the db is empty. If the db is empty the connector can't build the schema for the dataframe. Is this the case?

doc/docs/modules/ROOT/pages/streaming.adoc
```diff
@@ -0,0 +1,151 @@
...
+
+**ALL**: send all the data in the DB to the stream before reading new data
+|`NOW`
+|Yes^*^
+
+|===
+
+== Sink
+
+Writing a stream to Neo4j is pretty easy and can be done using any of the xref:writing.adoc#_write_data[3 write strategies].
```
moxious, line 55: add note here that we need some kind of link to basic spark streaming information.   The connector is the same, but spark streaming works very differently from the rest of spark batching.  Just some general link, "Here's background on Spark Streaming approach"

doc/docs/modules/ROOT/pages/streaming.adoc
```diff
@@ -0,0 +1,151 @@
...
+[source,scala]
+----
+import org.apache.spark.sql.{SaveMode, SparkSession}
+
+val spark = SparkSession.builder().getOrCreate()
+
+/**
+ * Example is made with kafka, but can be any Spark Streaming Dataframe
+ */
+val df = spark
```
moxious: if possible please do "python first" code examples

doc/docs/modules/ROOT/pages/streaming.adoc
utnaf: This is fixed by #387

doc/docs/modules/ROOT/pages/streaming.adoc
utnaf: @moxious ping on this one

doc/docs/modules/ROOT/pages/streaming.adoc
utnaf: > What does this mean? This doesn't do anything sensible, but wondering what we should say about this in the docs, if anything.
> 
> Notice the query doesn't include lastUpdated. How do queries intersect with streaming.property.name? I'm not sure....
> 
> Screen Shot 2021-09-14 at 2 03 20 PM

On this one we might state that the user need to take care of the the WHERE clause. Because usually for the `query` mode the connector doesn't interfere with the query provided by the user... except for the SKIP LIMIT part. Thoughs?

doc/docs/modules/ROOT/pages/streaming.adoc
moxious: resolving because I can't repro at this point either, I might have been in error, sorry

doc/docs/modules/ROOT/pages/streaming.adoc
moxious: probably need one small paragraph that calls out the "query" mode.  Say that in the case of query mode, the polling query is provided by the user (obviously) and so the connector disregards the "streaming.property.name" parameter in this case.   The way the docs are structured, informs the user to use this property, but this is a special case.  Maybe the parameter just doesn't apply there.

utnaf:
updated!

@conker84 @moxious I added some more detailed info and additional resources. I deployed this branch on our server.

conker84:
@utnaf thank you so much!

PR #363, opened 2021-05-30
Fixes #361: Change constraint naming convention

### LINKED ISSUE #361: Change constraint naming convention
If we have one dataframe with two columns [id:nullable,name:nullable] and two spark jobs like:

Job 1:
```scala
ds.write
      .format(classOf[DataSource].getName)
      .option("labels", ":City")
      .option("node.keys", "id")
      .option("schema.optimization.type", "NODE_CONSTRAINTS")
      .save()
```

Job 2:
```scala
ds.write
      .format(classOf[DataSource].getName)
      .option("labels", ":City")
      .option("node.keys", "name")
      .option("schema.optimization.type", "NODE_CONSTRAINTS")
      .save()
```

Under the hood we try to create two constraints:

Job 1:
```
"CREATE CONSTRAINT `spark_NODE_CONSTRAINTS_City`
ON (n:City) ASSERT (n.id) IS UNIQUE
```

Job 2:
```
"CREATE CONSTRAINT `spark_NODE_CONSTRAINTS_City`
ON (n:City) ASSERT (n.name) IS UNIQUE
```

The second constant will not be created because of the same name, we need to add the property into the constraint naming convention.

utnaf:
Fixes #361

### DIFF
--- a/common/src/main/scala/org/neo4j/spark/service/SchemaService.scala
+++ b/common/src/main/scala/org/neo4j/spark/service/SchemaService.scala
@@ -467,6 +467,7 @@ class SchemaService(private val options: Neo4jOptions, private val driverCache: 
         val quotedProps = props
           .map(prop => s"${Neo4jUtil.NODE_ALIAS}.${prop.quote()}")
           .mkString(", ")
+        val dashSeparatedProps = props.mkString("-")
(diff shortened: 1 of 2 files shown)


conker84 (approved):
Thank you @utnaf

PR #367, opened 2021-06-07
Only run docs workflow when something changed in doc folder

ggrossetie:
## Proposed Changes

  - Use [`on.<push|pull_request>.paths`](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestpaths) to run docs workflow when at least one path matches a pattern in the `paths` filter

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -2,10 +2,14 @@ name: Docs
 
 on:
   push:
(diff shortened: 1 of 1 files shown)


conker84 (approved):

conker84:
Thank you @Mogztter

PR #368, opened 2021-06-07
Disable lint:links task for now

ggrossetie:
## Proposed Changes

- Disable `npm run lint:links` task from "Docs" workflow. The task is unreliable and slow(ish). Enable it again when we found a solution.


conker84:
Thank you @Mogztter

PR #377, opened 2021-06-24
Fixes #369: add parameters to filters

### LINKED ISSUE #369: Values in filters should be set via parameters
Right now the values in the filters are written directly in the query:

```
MATCH (n:`Test4_Movie`) WHERE n.timestamp > 1623336494510 RETURN n
```

```
MATCH (source:`Test2_Person`) 
MATCH (target:`Test2_Post`) 
MATCH (source)-[rel:`LIKES`]->(target) 
WHERE rel.timestamp > 1623336511332 RETURN rel, source AS source, target AS target
```

When they actually should be passed as parameters to the query.

utnaf:
This handles the following cases:

* `WHERE age > 10 AND age < 20` can't have a parameter with the same name for both age values
* OR and AND are particular filters that contains other fillers (or.left, or.right), and each filter can be either an OR or an AND
* parameters name with a dot inside
* count query on streaming read (added parameters to count query since it has `timestamp`)
* quoted attribues
* `WHERE date > date($dateParam)`
* `$stream.offset` parameter

### DIFF
--- a/common/src/main/scala/org/neo4j/spark/reader/BasePartitionReader.scala
+++ b/common/src/main/scala/org/neo4j/spark/reader/BasePartitionReader.scala
@@ -34,10 +34,6 @@ abstract class BasePartitionReader(private val options: Neo4jOptions,
     Neo4jUtil.paramsFromFilters(filters)
       .foreach(p => params.put(p._1, p._2))
 
-    if (log.isDebugEnabled) {
-      log.debug(s"Query Parameters are: $params")
-    }
-
     params.asJava
   }
 
@@ -48,7 +44,13 @@ abstract class BasePartitionReader(private val options: Neo4jOptions,
       session = driverCache.getOrCreate().session(options.session.toNeo4jSession)
       transaction = session.beginTransaction()
       log.info(s"Running the following query on Neo4j: $query")
-      result = transaction.run(query, Values.value(getQueryParameters))
+
+      val queryParams = getQueryParameters
+      if (log.isDebugEnabled) {
+        log.debug(s"Query Parameters are: $queryParams")
+      }
+
+      result = transaction.run(query, Values.value(queryParams))
         .asScala
     }
 
--- a/common/src/main/scala/org/neo4j/spark/reader/BasePartitionReader.scala
+++ b/common/src/main/scala/org/neo4j/spark/reader/BasePartitionReader.scala
@@ -28,6 +28,19 @@ abstract class BasePartitionReader(private val options: Neo4jOptions,
   private val query: String = new Neo4jQueryService(options, new Neo4jQueryReadStrategy(filters, partitionSkipLimit, requiredColumns.fieldNames))
     .createQuery()
 
+  private lazy val values = {
+    val params = mutable.HashMap[String, Any]()
+    params.put(Neo4jQueryStrategy.VARIABLE_SCRIPT_RESULT, scriptResult)
+    Neo4jUtil.paramsFromFilters(filters)
+      .foreach(p => params.put(p._1, p._2))
+
+    if (log.isDebugEnabled) {
+      log.debug(s"Query Parameters are: $params")
+    }
+
+    params.asJava
+  }
+
   private val mappingService = new MappingService(new Neo4jReadMappingStrategy(options, requiredColumns), options)
 
   def next: Boolean = {
@@ -50,16 +63,5 @@ abstract class BasePartitionReader(private val options: Neo4jOptions,
     driverCache.close()
   }
 
-  protected def getQueryParameters: util.Map[String, Any] = {
-    val params = mutable.HashMap[String, Any]()
-    params.put(Neo4jQueryStrategy.VARIABLE_SCRIPT_RESULT, scriptResult)
-    Neo4jUtil.paramsFromFilters(filters)
-      .foreach(p => params.put(p._1, p._2))
-
-    if (log.isDebugEnabled) {
-      log.debug(s"Query Parameters are: $params")
-    }
-
-    params.asJava
-  }
+  protected def getQueryParameters: util.Map[String, Any] = values
 }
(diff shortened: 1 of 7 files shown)


common/src/main/scala/org/neo4j/spark/util/Neo4jUtil.scala
```diff
@@ -239,46 +242,111 @@ object Neo4jUtil {
 
   def connectorVersion: String = properties.getOrDefault("version", "UNKNOWN").toString
 
-  def valueToCypherExpression(value: Any): Expression = {
+  def getCorrectProperty(container: PropertyContainer, attribute: String): Property = {
+    container.property(attribute.split('.'): _*)
+  }
+
+  @tailrec
```
conker84: why this?

common/src/main/scala/org/neo4j/spark/util/Neo4jUtil.scala
```diff
@@ -239,46 +242,111 @@ object Neo4jUtil {
...
+      case _ => value
     }
   }
 
-  def getCorrectProperty(container: PropertyContainer, attribute: String): Property = {
-    container.property(attribute.split('.') : _*)
+  /**
+   * df: we need this to handle scenarios like `WHERE age > 19 and age < 22`,
+   * so we can't basically add a parameter named $age.
+   * So we base64 encode the value to ensure a unique parameter name
```
conker84: great idea!

common/src/main/scala/org/neo4j/spark/util/Neo4jUtil.scala
```diff
@@ -239,46 +242,111 @@ object Neo4jUtil {
...
+    val attributeValue = if (value == null) {
+      "NULL"
+    }
+    else {
+      value.toString
+    }
+
+    val base64ed = java.util.Base64.getEncoder.encodeToString(attributeValue.getBytes())
+
+    s"${base64ed}_${attribute.unquote()}".quote()
```
conker84: as above please move it into an implicit method

common/src/test/scala/org/neo4j/spark/service/Neo4jQueryServiceTest.scala
```diff
@@ -260,9 +282,12 @@ class Neo4jQueryServiceTest {
 
     val query: String = new Neo4jQueryService(neo4jOptions, new Neo4jQueryReadStrategy(filters)).createQuery()
 
+    val paramOneName = "$" + Neo4jUtil.createParameterName("source.name", "John Doe")
+    val paramTwoName = "$" + Neo4jUtil.createParameterName("target.name", "John Doe")
+
     assertEquals("MATCH (source:`Person`) " +
       "MATCH (target:`Person`) " +
-      "MATCH (source)-[rel:`KNOWS`]->(target) WHERE (source.name = 'John Doe' OR target.name = 'John Doe') RETURN rel, source AS source, target AS target", query)
+      s"MATCH (source)-[rel:`KNOWS`]->(target) WHERE (source.name = $paramOneName OR target.name = $paramTwoName) RETURN rel, source AS source, target AS target", query)
```
conker84, line 291: as above

common/src/main/scala/org/neo4j/spark/util/Neo4jUtil.scala
```diff
@@ -239,46 +242,111 @@ object Neo4jUtil {
...
+      case f: LessThan => Seq(createParameterName(f.attribute, f.value), f.value)
+      case f: LessThanOrEqual => Seq(createParameterName(f.attribute, f.value), f.value)
+      case f: In => Seq(createParameterName(f.attribute, f.values), f.values)
+      case f: StringStartsWith => Seq(createParameterName(f.attribute, f.value), f.value)
+      case f: StringEndsWith => Seq(createParameterName(f.attribute, f.value), f.value)
+      case f: StringContains => Seq(createParameterName(f.attribute, f.value), f.value)
+      case f: Not => getAttributeAndValue(f.child)
+      case _ => Seq()
+    }
+  }
```
conker84: maybe it's better to transform to an implicit function

conker84 (changes requested):

common/src/main/scala/org/neo4j/spark/reader/BasePartitionReader.scala
```diff
@@ -59,4 +50,16 @@ abstract class BasePartitionReader(private val options: Neo4jOptions,
...
+    val params = mutable.HashMap[String, Any]()
+    params.put(Neo4jQueryStrategy.VARIABLE_SCRIPT_RESULT, scriptResult)
+    Neo4jUtil.paramsFromFilters(filters)
+      .foreach(p => params.put(p._1, p._2))
+
+    if (log.isDebugEnabled) {
+      log.debug(s"Query Parameters are: $params")
+    }
+
+    params.asJava
```
conker84: please store `params.asJava` in a lazy val

conker84 (changes requested):

common/src/main/scala/org/neo4j/spark/util/Neo4jUtil.scala
utnaf: optimizes the compilation for tail recursion https://www.scala-lang.org/api/2.12.1/scala/annotation/tailrec.html

common/src/main/scala/org/neo4j/spark/util/Neo4jUtil.scala
utnaf: you mean the `createParameterName` ?

conker84: I meant both!

common/src/main/scala/org/neo4j/spark/util/Neo4jImplicits.scala
```diff
@@ -112,6 +139,22 @@ object Neo4jImplicits {
     }
 
     def getAttributeWithoutEntityName: Option[String] = filter.getAttribute.map(_.unquote().split('.').tail.mkString("."))
+
+    def getAttributeAndValue: Seq[Any] = {
```
conker84, line 149: why don't we manage EqualNullSafe and And/Or?

utnaf: And/Or are handled by [this function](https://github.com/neo4j-contrib/neo4j-spark-connector/pull/377/files#diff-47b5cfa4eae29a6ccac76fcffed8b50a0851e27835665551374237611d31f76aR97-R103) called [here](https://github.com/neo4j-contrib/neo4j-spark-connector/pull/377/files#diff-7efd6a0c28a95713ded5f486215a2d67baca6402591f2ce6423ace245de612a1R249-R252). I added a comment to explain better.

Added EqualNullSafe case.

conker84 (approved):
LGTM

conker84:
thank you @utnaf

PR #387, opened 2021-09-20
fixes #386: Streaming Source dups record in Databricks env

### LINKED ISSUE #386: Streaming Source dups record in Databricks env
Streaming Source dups record in Databricks env, as explained by @moxious here:

https://github.com/neo4j-contrib/neo4j-spark-connector/pull/346#discussion_r708169710

conker84:
Fixes #386 

The current implementation doesn't allow to return the last read offset from the Executors to the Driver, we now use accumulators in order to share variables between Executors to the Driver.

 - Fixed the bug
 - This behaviour came out in Databricks env where executors are in different JVMs so it's not easily reproducible with tests, we can think about starting a spark cluster within docker containers in the future...


common/src/main/scala/org/neo4j/spark/streaming/OffsetStorage.scala
```diff
@@ -1,28 +1,194 @@
...
+
+  override def close(): Unit = synchronized {
+    var session: Session = null
+    try {
+      session = driverCache.getOrCreate().session(neo4jOptions.session.toNeo4jSession())
+      session.writeTransaction(new TransactionWork[Unit] {
+        override def execute(tx: Transaction): Unit = {
+          tx.run(
+            s"""
+               |MERGE (n:${Neo4jAccumulator.LABEL}{${Neo4jAccumulator.KEY}: ${'$'}jobId})
```
utnaf, line 105: Is this because forcing the write transaction will get us straight to the leader? I would add a comment here too

common/src/main/scala/org/neo4j/spark/streaming/OffsetStorage.scala
```diff
@@ -1,28 +1,194 @@
...
+  extends OffsetStorage[lang.Long, lang.Long] {
+
+  private lazy val driverCache = new DriverCache(neo4jOptions.connection, jobId)
+  add(initialValue)
 
-  private val cache = new ConcurrentHashMap[String, java.lang.Long]()
+  override def copy(): AccumulatorV2[lang.Long, lang.Long] = {
+    val copy = new Neo4jAccumulator(neo4jOptions, jobId)
+    copy
+  }
```
utnaf:
```suggestion
  override def copy(): AccumulatorV2[lang.Long, lang.Long] = new Neo4jAccumulator(neo4jOptions, jobId)
```

utnaf (commented):
Overall it looks good. I wonder if it's worth to write a doc paragraph about the differences between Spark 2.4 and Spark 3

PR #390, opened 2021-10-29
Read JAR version from pom.xml in Python CI

utnaf:
Reading the jar version from the pom and use it in Python CI for the build command.


conker84:
Thank you so much @utnaf

PR #394, opened 2021-11-18
Add playbooks for docs build processes

recrwplay:
Adds two playbooks:
- preview.yml - for local builds, to generate HTML from locally checked out content in the current branch
- publish.yml - for CI builds, to generate HTML direct from the git branches, for all released versions


davidoliverSP2 (approved):
Many thanks @recrwplay 

👍

PR #398, opened 2021-12-17
4.1 renaming aura

NataliaIvakina:
Neo4j Aura has become Neo4j AuraDB, so we need to make the corresponding revisions in the manual.


conker84:
@NataliaIvakina thank you!

PR #418, opened 2022-03-16
Upgrade docs to Antora 3

recrwplay:
- Updates packages for Antora 3
- Updates publish.yml playbook

PR #425, opened 2022-04-06
Update docs.yml

conker84:
After #418 the action broke so we need to update it


recrwplay (approved):
Looks good. Apologies - I didn't realise you had actions that depended on a script I was removing.

conker84:
No worries, thank you so much @recrwplay !

PR #461, opened 2023-04-17
Update Neo4j versions used by integration tests

fbiville:
This sets the latest 4.x and latest 5.x Neo4j versions.

This only affects container-based tests.


ali-ince (approved):
looks good to me.

PR #463, opened 2023-04-17
Make the build script optional - use sensible Maven defaults

fbiville:
This now plays nicer with IDEs (and newcomers like me ^^) by making the build script optional to use locally.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -211,6 +211,26 @@
                     </execution>
                 </executions>
             </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-enforcer-plugin</artifactId>
+                <version>3.3.0</version>
+                <executions>
+                    <execution>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -93,6 +96,9 @@
         </profile>
         <profile>
             <id>neo4j-5</id>
+            <activation>
+                <activeByDefault>true</activeByDefault>
+            </activation>
             <properties>
                 <neo4j.version>5.4.0</neo4j.version>
```
ali-ince, line 103: should we also update this to be `5.6.0` or soon `5.7.0`?

fbiville: I think so, I reduced the PR to the bare minimum but happy to include this as well

fbiville: actually, I forgot I did that here: https://github.com/neo4j-contrib/neo4j-spark-connector/pull/461/ ^^

ali-ince (approved):

PR #466, opened 2023-04-18
Remove always-false comparison in Spark version checker

fbiville:
Caught via a compiler warning

### DIFF
--- a/common/src/main/scala/org/neo4j/spark/util/Validations.scala
+++ b/common/src/main/scala/org/neo4j/spark/util/Validations.scala
@@ -53,7 +53,7 @@ case class ValidateSparkVersion(supportedVersions: String*) extends Validation {
     val splittedVersion = sparkVersion.split("\\.")
 
     ValidationUtil.isTrue(
-      supportedVersions
+      sparkVersion == "UNKNOWN" || supportedVersions
         .flatMap(_.split("\\.").zip(splittedVersion))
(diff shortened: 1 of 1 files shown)


common/src/main/scala/org/neo4j/spark/util/Validations.scala
```diff
@@ -53,7 +53,7 @@ case class ValidateSparkVersion(supportedVersions: String*) extends Validation {
     val splittedVersion = sparkVersion.split("\\.")
 
     ValidationUtil.isTrue(
-      sparkVersion(0) == "UNKNOWN" || supportedVersions
```
fbiville, line 56: this was comparing a char with a string

common/src/main/scala/org/neo4j/spark/util/Validations.scala
```diff
@@ -53,7 +53,7 @@ case class ValidateSparkVersion(supportedVersions: String*) extends Validation {
...
+      supportedVersions
```
conker84: I think that maybe we can improve it in this way.
```suggestion
      sparkVersion == "UNKNOWN" || supportedVersions
```

If for some reason we cannot retrieve the version we don't want to block the user.

conker84 (commented):
Thank you so much @fbiville please check my comment

conker84 (approved):

conker84:
@fbiville thank you so much!

PR #492, opened 2023-05-12
Support push down v2 filters

fbiville:
... internally propagate them as v1 filters, since these are already supported and easier to work with

### DIFF
--- a/common/src/main/scala/org/neo4j/spark/util/Neo4jUtil.scala
+++ b/common/src/main/scala/org/neo4j/spark/util/Neo4jUtil.scala
@@ -157,7 +157,7 @@ object Neo4jUtil {
 
   def convertFromSpark(value: Any, dataType: DataType = null): Value = value match {
     case date: java.sql.Date => convertFromSpark(date.toLocalDate, dataType)
-    case timestamp: java.sql.Timestamp => convertFromSpark(timestamp.toInstant.atZone(ZoneOffset.UTC), dataType)
+    case timestamp: java.sql.Timestamp => convertFromSpark(timestamp.toLocalDateTime, dataType)
     case intValue: Int if dataType != null && dataType == DataTypes.DateType => convertFromSpark(DateTimeUtils
       .toJavaDate(intValue), dataType)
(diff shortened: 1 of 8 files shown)


common/src/main/scala/org/neo4j/spark/util/Neo4jUtil.scala
```diff
@@ -260,7 +260,7 @@ object Neo4jUtil {
   def toParamValue(value: Any): Any = {
     value match {
       case date: java.sql.Date => date.toString
```
fbiville, line 251: I'm wondering why this line did not need to change. Is it covered by tests?

common/src/main/scala/org/neo4j/spark/util/Neo4jUtil.scala
```diff
@@ -155,23 +155,23 @@ object Neo4jUtil {
...
-    case date: java.sql.Date => convertFromSpark(date.toLocalDate, schema)
-    case timestamp: java.sql.Timestamp => convertFromSpark(timestamp.toInstant.atZone(ZoneOffset.UTC), schema)
-    case intValue: Int if schema != null && schema.dataType == DataTypes.DateType => convertFromSpark(DateTimeUtils
-      .toJavaDate(intValue), schema)
-    case longValue: Long if schema != null && schema.dataType == DataTypes.TimestampType => convertFromSpark(DateTimeUtils
-      .toJavaTimestamp(longValue), schema)
+  def convertFromSpark(value: Any, dataType: DataType = null): Value = value match {
+    case date: java.sql.Date => convertFromSpark(date.toLocalDate, dataType)
+    case timestamp: java.sql.Timestamp => convertFromSpark(timestamp.toLocalDateTime, dataType)
+    case intValue: Int if dataType != null && dataType == DataTypes.DateType => convertFromSpark(DateTimeUtils
```
fbiville: we can simplify these checks to just rely on `==`

fbiville: fixed in e5c9ad496cb5791b049f9426d3c97ee8cac85d19

conker84 (approved):
LGTM Thank you @fbiville !

PR #499, opened 2023-05-22
Fix spark version validation on EMR

liurong79:


conker84:
Thank you so much for the contribution, unfortunately we solved it with #501 . I'm closing this

PR #513, opened 2023-06-05
Broken links fix

lidiazuin:
The Values and types section was moved in the Cypher manual, thus the broken links being fixed here. There might be the need to cherry-pick this backwards.


fbiville (approved):
Thanks!

PR #520, opened 2023-06-13
Implement pushdown limit support

fbiville:
Fixes #519

PR #526, opened 2023-06-26
Implement Top N pushdown

fbiville:
Stacked on top of https://github.com/neo4j-contrib/neo4j-spark-connector/pull/520

### DIFF
--- a/common/src/test/scala/org/neo4j/spark/util/DummyNamedReference.scala
+++ b/common/src/test/scala/org/neo4j/spark/util/DummyNamedReference.scala
@@ -1,0 +1,7 @@
+package org.neo4j.spark.util
+
+import org.apache.spark.sql.connector.expressions.NamedReference
+
+abstract class DummyNamedReference extends NamedReference {
+  override def toString: String = describe()
+}
--- a/common/src/test/scala/org/neo4j/spark/util/DummyNamedReference.scala
+++ b/common/src/test/scala/org/neo4j/spark/util/DummyNamedReference.scala
@@ -2,6 +2,9 @@ package org.neo4j.spark.util
 
 import org.apache.spark.sql.connector.expressions.NamedReference
 
-abstract class DummyNamedReference extends NamedReference {
+class DummyNamedReference(private val fields: String*) extends NamedReference {
+  override def fieldNames(): Array[String] = fields.toArray
+
+  override def describe(): String = fields.mkString(", ")
   override def toString: String = describe()
 }
--- a/spark-3/src/main/scala/org/neo4j/spark/reader/Neo4jScanBuilder.scala
+++ b/spark-3/src/main/scala/org/neo4j/spark/reader/Neo4jScanBuilder.scala
@@ -1,11 +1,12 @@
 package org.neo4j.spark.reader
 
 import org.apache.spark.internal.Logging
-import org.apache.spark.sql.connector.expressions.Expression
+import org.apache.spark.sql.connector.expressions.{Expression, SortOrder}
 import org.apache.spark.sql.connector.expressions.aggregate.{AggregateFunc, Aggregation}
 import org.apache.spark.sql.connector.expressions.filter.Predicate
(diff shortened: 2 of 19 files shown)


spark-3/src/main/scala/org/neo4j/spark/reader/Neo4jScanBuilder.scala
```diff
@@ -126,4 +126,7 @@ class Neo4jScanBuilder(neo4jOptions: Neo4jOptions, jobId: String, schema: Struct
     topN = Some(TopN(limit, orders))
     true
   }
+
+  // otherwise doesn't compile in Scala 2.12
+  override def isPartiallyPushed: Boolean = true
```
fbiville, line 131: so the interface is different between spark for Scala 2.12 and 2.13??

common/src/test/scala/org/neo4j/spark/util/DummyNamedReference.scala
```diff
@@ -0,0 +1,7 @@
+package org.neo4j.spark.util
+
+import org.apache.spark.sql.connector.expressions.NamedReference
+
+abstract class DummyNamedReference extends NamedReference {
```
fbiville: maybe we could make it a concrete class with a constructor that takes a list of strings and `describe` just joins the list into a string? that would make the tests much less verbose. WDYT?
Happy to do it if you agree with the change

spark-3/src/main/scala/org/neo4j/spark/reader/Neo4jScanBuilder.scala
conker84: I don't think so, I just get this error by the compiler in Intellj:

```
class Neo4jScanBuilder inherits conflicting members:
  method isPartiallyPushed in trait SupportsPushDownLimit of type ()Boolean  and
  method isPartiallyPushed in trait SupportsPushDownTopN of type ()Boolean
(Note: this can be resolved by declaring an override in class Neo4jScanBuilder.)
class Neo4jScanBuilder(neo4jOptions: Neo4jOptions, jobId: String, schema: StructType)
```

common/src/test/scala/org/neo4j/spark/util/DummyNamedReference.scala
conker84: yes I did it as is as was easier to do a find/replace with just the name of the class, but I can do it in that way

conker84: done

spark-3/src/main/scala/org/neo4j/spark/reader/Neo4jScanBuilder.scala
conker84: As a side note, both of them are default methods that return `true`.

conker84 (approved):

conker84 (approved):

fbiville:
PR is now rebased and the doc changes have moved to https://github.com/neo4j/docs-spark/pull/16

spark-3/src/test/scala/org/neo4j/spark/DataSourceReaderTSE.scala
```diff
@@ -1770,6 +1771,37 @@ class DataSourceReaderTSE extends SparkConnectorScalaBaseTSE {
...
+
+    val df = ss.read
+      .format(classOf[DataSource].getName)
+      .option("url", SparkConnectorScalaSuiteIT.server.getBoltUrl)
+      .option("relationship", "BOUGHT")
+      .option("relationship.source.labels", "Person")
+      .option("relationship.target.labels", "Product")
+      .load
+      .select("`target.name`", "`target.id`")
+      .orderBy(col("`target.name`").desc) // FIXME (generated query is wrong)
```
ali-ince: is this `FIXME` just forgotten or is there something that we really need to fix?

fbiville: ah wait, I think we forgot about it. There was a test case where it was causing issues

fbiville:
Now rebased again.

PR #539, opened 2023-08-23
Update Scala 2.13 version

fbiville:


pom.xml
```diff
@@ -80,7 +80,7 @@
                 <activeByDefault>true</activeByDefault>
             </activation>
             <properties>
-                <scala.version>2.13.4</scala.version>
+                <scala.version>2.13.9</scala.version>
```
fbiville, line 83: maybe we merge just this change @conker84, wdyt?

conker84: sure

PR #540, opened 2023-08-23
added fixed bug in version validation

conker84:
The test `testThrowsExceptionSparkVersionIsNotSupported` was a false negative because it was supposed to be valid. In order to check the problem you can just update the Spark version from 3.3.2 to 3.4.2 and magically it's a valid version.

I fixed the problem but also changed the class name to a more meaningful one.


spark-3/src/test/scala/org/neo4j/spark/VersionValidationTest.scala
```diff
@@ -14,7 +14,7 @@ class VersionValidationTest extends SparkConnectorScalaBaseTSE {
       .map { _.version }
       .getOrElse("UNKNOWN")
     try {
-      Validations.validate(ValidateSparkVersion("2.4"))
```
conker84, line 17: This should be valid also in version 3.3.2

fbiville (approved):

PR #541, opened 2023-08-24
added driver fetch size

conker84:
added driver fetch size


common/src/main/scala/org/neo4j/spark/util/Neo4jOptions.scala
```diff
@@ -301,6 +303,7 @@ case class Neo4jDriverOptions(
     if (acquisitionTimeout > -1) builder.withConnectionAcquisitionTimeout(acquisitionTimeout, TimeUnit.MILLISECONDS)
     if (livenessCheckTimeout > -1) builder.withConnectionLivenessCheckTimeout(livenessCheckTimeout, TimeUnit.MILLISECONDS)
     if (connectionTimeout > -1) builder.withConnectionTimeout(connectionTimeout, TimeUnit.MILLISECONDS)
+    if (fetchSize > -1) builder.withFetchSize(fetchSize)
```
fbiville: beware that `-1` is actually a valid value for the Java driver.
> By default, the value is set to 1000. Use -1 to disables back pressure and config client to pull all records at once after each run.

conker84: Done

fbiville (approved):

fbiville:
Let's close this PR for now. We can reopen it if we ever think it makes sense to have it.

conker84:
@fbiville Do you think that cannot be useful for heavy operations? As it's basically ready, for me is just an added value to the connector.

fbiville:
Let's discuss this internally and re-evaluate.
There could be value indeed, but no-one has asked for it yet and expanding the configuration surface means more choice and potentially more confusion for the end user too.

PR #551, opened 2023-09-28
Update Maven wrapper

fbiville:
This fixes a relative path traversal vulnerability (CWE-23).


ali-ince (approved):
looks good, latest maven could be a good move too.

PR #566, opened 2023-11-22
feat: add metrics support to the sink path

ali-ince:


common/src/main/scala/org/neo4j/spark/writer/DataWriterMetrics.scala
```diff
@@ -0,0 +1,140 @@
...
+      DataWriterMetric(PROPERTIES_SET, propertiesSet.longValue()),
+      DataWriterMetric(LABELS_ADDED, labelsAdded.longValue()),
+      DataWriterMetric(LABELS_REMOVED, labelsRemoved.longValue())
+    ).toArray
+  }
+
+}
+
+object DataWriterMetrics {
+  final val RECORDS_WRITTEN = "recordsWritten"
```
venikkin: I don't know that are the naming conventions spark have for metrics, but looking at existing (for instance https://spark.apache.org/docs/latest/monitoring.html#executor-metrics), spark sometimes combines them by groups like `memoryMetrics.*` or `peakMemoryMetrics`. I wonder if it makes sense to prefix our custom metrics with something like `neo4jMetrics.*`?

venikkin (approved):
Looks good. Have small suggestion / question reg the metrics naming.

common/src/main/scala/org/neo4j/spark/writer/DataWriterMetrics.scala
fbiville: I like this idea. The only potential gotcha is that the name of custom metrics that seem to surface are based on the descriptions we provide, not the name.
WDYT @ali-ince?

ali-ince: Yeah, I agree that it's a nice idea. Let's use namespaced naming scheme as @venikkin suggests, and we can update the descriptions in the future if we need to.

fbiville: sorry, a bit slow on that one, but here is the change: c9d214d5f8f325661d34b4944f5bddb5b5a9190a.
what do you both think?

venikkin (approved):

PR #581, opened 2023-12-18
Inferred schema type constraints + improved constraints management

conker84:
- [x] Added schema type constraints
- [x] Improved constraints management 

The docs are in this PR: https://github.com/neo4j/docs-spark/pull/17

### DIFF
--- a/common/src/main/scala/org/neo4j/spark/converter/DataConverter.scala
+++ b/common/src/main/scala/org/neo4j/spark/converter/DataConverter.scala
@@ -1,0 +1,181 @@
+package org.neo4j.spark.converter
+
+import org.apache.spark.sql.catalyst.InternalRow
+import org.apache.spark.sql.catalyst.expressions.{GenericRowWithSchema, UnsafeArrayData, UnsafeMapData, UnsafeRow}
+import org.apache.spark.sql.catalyst.util.{ArrayBasedMapData, ArrayData, DateTimeUtils}
+import org.apache.spark.sql.types._
+import org.apache.spark.unsafe.types.UTF8String
+import org.neo4j.driver.internal._
+import org.neo4j.driver.{Value, Values}
+import org.neo4j.spark.service.SchemaService
+import org.neo4j.spark.util.Neo4jUtil
+
+import java.time._
+import java.time.format.DateTimeFormatter
+import scala.annotation.tailrec
+import scala.collection.JavaConverters._
+
+trait DataConverter[T] {
+  def convert(value: Any, dataType: DataType = null): T
+
+  @tailrec
+  private[converter] final def extractStructType(dataType: DataType): StructType = dataType match {
+    case structType: StructType => structType
+    case mapType: MapType => extractStructType(mapType.valueType)
+    case arrayType: ArrayType => extractStructType(arrayType.elementType)
+    case _ => throw new UnsupportedOperationException(s"$dataType not supported")
+  }
+}
+
+object SparkToNeo4jDataConverter {
+  def apply(): SparkToNeo4jDataConverter = new SparkToNeo4jDataConverter()
+}
+
+class SparkToNeo4jDataConverter extends DataConverter[Value] {
+  override def convert(value: Any, dataType: DataType): Value = {
+    value match {
+      case date: java.sql.Date => convert(date.toLocalDate, dataType)
+      case timestamp: java.sql.Timestamp => convert(timestamp.toLocalDateTime, dataType)
+      case intValue: Int if dataType == DataTypes.DateType => convert(DateTimeUtils
+        .toJavaDate(intValue), dataType)
+      case longValue: Long if dataType == DataTypes.TimestampType => convert(DateTimeUtils
+        .toJavaTimestamp(longValue), dataType)
+      case unsafeRow: UnsafeRow => {
+        val structType = extractStructType(dataType)
+        val row = new GenericRowWithSchema(unsafeRow.toSeq(structType).toArray, structType)
+        convert(row)
+      }
+      case struct: GenericRowWithSchema => {
+        def toMap(struct: GenericRowWithSchema): Value = {
+          Values.value(
+            struct.schema.fields.map(
+              f => f.name -> convert(struct.getAs(f.name), f.dataType)
+            ).toMap.asJava)
+        }
+
+        try {
+          struct.getAs[UTF8String]("type").toString match {
+            case SchemaService.POINT_TYPE_2D => Values.point(struct.getAs[Number]("srid").intValue(),
+              struct.getAs[Number]("x").doubleValue(),
+              struct.getAs[Number]("y").doubleValue())
+            case SchemaService.POINT_TYPE_3D => Values.point(struct.getAs[Number]("srid").intValue(),
+              struct.getAs[Number]("x").doubleValue(),
+              struct.getAs[Number]("y").doubleValue(),
+              struct.getAs[Number]("z").doubleValue())
+            case SchemaService.DURATION_TYPE => Values.isoDuration(struct.getAs[Number]("months").longValue(),
+              struct.getAs[Number]("days").longValue(),
+              struct.getAs[Number]("seconds").longValue(),
+              struct.getAs[Number]("nanoseconds").intValue())
+            case SchemaService.TIME_TYPE_OFFSET => Values.value(OffsetTime.parse(struct.getAs[UTF8String]("value").toString))
+            case SchemaService.TIME_TYPE_LOCAL => Values.value(LocalTime.parse(struct.getAs[UTF8String]("value").toString))
+            case _ => toMap(struct)
+          }
+        } catch {
+          case _: Throwable => toMap(struct)
+        }
+      }
+      case unsafeArray: UnsafeArrayData => {
+        val sparkType = dataType match {
+          case arrayType: ArrayType => arrayType.elementType
+          case _ => dataType
+        }
+        val javaList = unsafeArray.toSeq[AnyRef](sparkType)
+          .map(elem => convert(elem, sparkType))
+          .asJava
+        Values.value(javaList)
+      }
+      case unsafeMapData: UnsafeMapData => { // Neo4j only supports Map[String, AnyRef]
+        val mapType = dataType.asInstanceOf[MapType]
+        val map: Map[String, AnyRef] = (0 until unsafeMapData.numElements())
+          .map(i => (unsafeMapData.keyArray().getUTF8String(i).toString, unsafeMapData.valueArray().get(i, mapType.valueType)))
+          .toMap[String, AnyRef]
+          .mapValues(innerValue => convert(innerValue, mapType.valueType))
+          .toMap[String, AnyRef]
+        Values.value(map.asJava)
+      }
+      case string: UTF8String => convert(string.toString)
+      case _ => Values.value(value)
+    }
+  }
+}
+
+object Neo4jToSparkDataConverter {
+  def apply(): Neo4jToSparkDataConverter = new Neo4jToSparkDataConverter()
+}
+
+class Neo4jToSparkDataConverter extends DataConverter[Any] {
+  override def convert(value: Any, dataType: DataType): Any = {
+    if (dataType != null && dataType == DataTypes.StringType && value != null && !value.isInstanceOf[String]) {
+      convert(Neo4jUtil.mapper.writeValueAsString(value), dataType)
+    } else {
+      value match {
+        case node: InternalNode => {
+          val map = node.asMap()
+          val structType = extractStructType(dataType)
+          val fields = structType
+            .filter(field => field.name != Neo4jUtil.INTERNAL_ID_FIELD && field.name != Neo4jUtil.INTERNAL_LABELS_FIELD)
+            .map(field => convert(map.get(field.name), field.dataType))
+          InternalRow.fromSeq(Seq(convert(node.id()), convert(node.labels())) ++ fields)
+        }
+        case rel: InternalRelationship => {
+          val map = rel.asMap()
+          val structType = extractStructType(dataType)
+          val fields = structType
+            .filter(field => field.name != Neo4jUtil.INTERNAL_REL_ID_FIELD
+              && field.name != Neo4jUtil.INTERNAL_REL_TYPE_FIELD
+              && field.name != Neo4jUtil.INTERNAL_REL_SOURCE_ID_FIELD
+              && field.name != Neo4jUtil.INTERNAL_REL_TARGET_ID_FIELD)
+            .map(field => convert(map.get(field.name), field.dataType))
+          InternalRow.fromSeq(Seq(convert(rel.id()),
+            convert(rel.`type`()),
+            convert(rel.startNodeId()),
+            convert(rel.endNodeId())) ++ fields)
+        }
+        case d: InternalIsoDuration => {
+          val months = d.months()
+          val days = d.days()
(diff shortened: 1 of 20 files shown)


common/src/main/scala/org/neo4j/spark/converter/DataConverter.scala
```diff
@@ -0,0 +1,181 @@
...
+class SparkToNeo4jDataConverter extends DataConverter[Value] {
+  override def convert(value: Any, dataType: DataType): Value = {
+    value match {
+      case date: java.sql.Date => convert(date.toLocalDate, dataType)
+      case timestamp: java.sql.Timestamp => convert(timestamp.toLocalDateTime, dataType)
+      case intValue: Int if dataType == DataTypes.DateType => convert(DateTimeUtils
+        .toJavaDate(intValue), dataType)
+      case longValue: Long if dataType == DataTypes.TimestampType => convert(DateTimeUtils
+        .toJavaTimestamp(longValue), dataType)
+      case unsafeRow: UnsafeRow => {
```
fbiville, line 43: maybe worth adding a comment with concrete examples where this `UnsafeRow` is useful

common/src/main/scala/org/neo4j/spark/converter/DataConverter.scala
```diff
@@ -0,0 +1,181 @@
...
+              struct.getAs[Number]("x").doubleValue(),
+              struct.getAs[Number]("y").doubleValue(),
+              struct.getAs[Number]("z").doubleValue())
+            case SchemaService.DURATION_TYPE => Values.isoDuration(struct.getAs[Number]("months").longValue(),
+              struct.getAs[Number]("days").longValue(),
+              struct.getAs[Number]("seconds").longValue(),
+              struct.getAs[Number]("nanoseconds").intValue())
+            case SchemaService.TIME_TYPE_OFFSET => Values.value(OffsetTime.parse(struct.getAs[UTF8String]("value").toString))
+            case SchemaService.TIME_TYPE_LOCAL => Values.value(LocalTime.parse(struct.getAs[UTF8String]("value").toString))
+            case _ => toMap(struct)
```
fbiville, line 72: does it cover the case where `type` is not defined?

common/src/main/scala/org/neo4j/spark/converter/DataConverter.scala
```diff
@@ -0,0 +1,181 @@
...
+          }
+        } catch {
+          case _: Throwable => toMap(struct)
```
fbiville, line 75: shouldn't we throw here?

common/src/main/scala/org/neo4j/spark/converter/DataConverter.scala
```diff
@@ -0,0 +1,181 @@
+package org.neo4j.spark.converter
+
+import org.apache.spark.sql.catalyst.InternalRow
+import org.apache.spark.sql.catalyst.expressions.{GenericRowWithSchema, UnsafeArrayData, UnsafeMapData, UnsafeRow}
```
fbiville: why do we need these `UnsafeXxx` types? that seems quite internal-ish?

common/src/main/scala/org/neo4j/spark/converter/DataConverter.scala
```diff
@@ -0,0 +1,181 @@
...
+  def apply(): Neo4jToSparkDataConverter = new Neo4jToSparkDataConverter()
+}
+
+class Neo4jToSparkDataConverter extends DataConverter[Any] {
+  override def convert(value: Any, dataType: DataType): Any = {
+    if (dataType != null && dataType == DataTypes.StringType && value != null && !value.isInstanceOf[String]) {
+      convert(Neo4jUtil.mapper.writeValueAsString(value), dataType)
+    } else {
+      value match {
+        case node: InternalNode => {
```
fbiville: do we need the concrete internal types here?
ain't the type outlined in https://neo4j.com/docs/java-reference/current/extending-neo4j/values-and-types/ enough?

common/src/main/scala/org/neo4j/spark/converter/TypeConverter.scala
```diff
@@ -0,0 +1,135 @@
...
+  private val mapping: Map[DataType, String] = Map(
+    DataTypes.BooleanType -> "BOOLEAN",
+    DataTypes.StringType -> "STRING",
+    DataTypes.IntegerType -> "INTEGER",
+    DataTypes.LongType -> "INTEGER",
+    DataTypes.FloatType -> "FLOAT",
+    DataTypes.DoubleType -> "FLOAT",
+    DataTypes.DateType -> "DATE",
+    DataTypes.TimestampType -> "LOCAL DATETIME",
+//    DataTypes.TimestampNTZType -> "ZONED DATETIME",
```
fbiville: maybe remove these?

common/src/main/scala/org/neo4j/spark/service/SchemaService.scala
```diff
@@ -621,57 +626,147 @@ class SchemaService(private val options: Neo4jOptions, private val driverCache:
...
+    val dashSeparatedProps = keys.values.mkString("-")
+    val constraintName = s"spark_${entityType}_${constraintType.replace(s"$entityType ", "")}-CONSTRAINT_${entityIdentifier}_$dashSeparatedProps".quote()
+    val props = keys.values.map(_.quote()).map("e." + _).mkString(", ")
+    val asciiRepresentation: String = createGraphEntityAsciiRepresentation(entityType, entityIdentifier)
+    session.writeTransaction(tx => {
+      tx.run(s"CREATE CONSTRAINT $constraintName IF NOT EXISTS FOR $asciiRepresentation REQUIRE ($props) IS $constraintType")
+    })
+  }
+
+  private def createGraphEntityAsciiRepresentation(entityType: String, entityIdentifier: String) = {
```
fbiville:
```suggestion
  private def createCypherPattern(entityType: String, entityIdentifier: String) = {
```
maybe?

common/src/main/scala/org/neo4j/spark/util/Validations.scala
```diff
@@ -45,6 +45,99 @@ case class ValidateSchemaOptions(neo4jOptions: Neo4jOptions, schema: StructType)
...
+        s"""You cannot combine `${Neo4jOptions.SCHEMA_OPTIMIZATION_TYPE}` with:
+           |- `${Neo4jOptions.SCHEMA_OPTIMIZATION_NODE_UNIQUE}`
+           |- `${Neo4jOptions.SCHEMA_OPTIMIZATION_NODE_KEY}`
+           |- `${Neo4jOptions.SCHEMA_OPTIMIZATION_RELATIONSHIP_UNIQUE}`
+           |- `${Neo4jOptions.SCHEMA_OPTIMIZATION_RELATIONSHIP_KEY}`
+           |- `${Neo4jOptions.SCHEMA_TYPE_CONSTRAINT}`
+      """)
+    }
+    if (schemaMetadata.optimization.nodeUnique && schemaMetadata.optimization.nodeKey) {
+      throw new IllegalArgumentException(
```
fbiville: this could be a warning instead (and then ignore the unique constraint)
but fine by me

common/src/main/scala/org/neo4j/spark/converter/DataConverter.scala
fbiville: and `InternalRow` is not even compatible with `Row`... 🤕

conker84: We're using these as they came from Spark, we are blocked here.

common/src/main/scala/org/neo4j/spark/converter/DataConverter.scala
conker84: yes

common/src/main/scala/org/neo4j/spark/converter/DataConverter.scala
conker84: I don't think that is the goal of this PR, I just moved the code to another class to simplify the management, if you want we can open an issue and discuss how to improve the code.

common/src/main/scala/org/neo4j/spark/converter/DataConverter.scala
conker84: changed here, in `InternalRelationship -> Relationship` and `InternalIsoDuration -> IsoDuration`

common/src/main/scala/org/neo4j/spark/util/Validations.scala
conker84: I usually prefer a fail-fast approach as otherwise it could lead to unexpected situations. So I would keep it as is

common/src/main/scala/org/neo4j/spark/service/SchemaService.scala
conker84: done

common/src/main/scala/org/neo4j/spark/converter/TypeConverter.scala
conker84: I left as reminder, because it's dependent from how we want to manage dates. If you want I can remove it

common/src/main/scala/org/neo4j/spark/converter/DataConverter.scala
fbiville: right, if it's existing code, then let's not change it

common/src/main/scala/org/neo4j/spark/converter/DataConverter.scala
fbiville: I gave it a quick try, and `InternalRow` is indeed required, but you can use `ArrayData` instead of `UnsafeArrayData` as well as `MapData` instead of `UnsafeMapData`.
That will at least reduce the coupling with internal types (not to 0, but still).

common/src/main/scala/org/neo4j/spark/converter/TypeConverter.scala
fbiville: let's open an issue instead, wdyt?

common/src/main/scala/org/neo4j/spark/converter/DataConverter.scala
conker84: Sure I also did for `GenericRowWithSchema -> GenericRow`

common/src/main/scala/org/neo4j/spark/converter/TypeConverter.scala
conker84: Done

common/src/main/scala/org/neo4j/spark/converter/TypeConverter.scala
```diff
@@ -0,0 +1,129 @@
...
+    DataTypes.StringType -> "STRING",
+    DataTypes.IntegerType -> "INTEGER",
+    DataTypes.LongType -> "INTEGER",
+    DataTypes.FloatType -> "FLOAT",
+    DataTypes.DoubleType -> "FLOAT",
+    DataTypes.DateType -> "DATE",
+    DataTypes.TimestampType -> "LOCAL DATETIME",
+    durationType -> "DURATION",
+    pointType -> "POINT",
+    // Cypher graph entities does not allow null values in arrays
```
fbiville:
```suggestion
    // Cypher graph entities do not allow null values in arrays
```

common/src/main/scala/org/neo4j/spark/service/MappingService.scala
```diff
@@ -70,8 +73,8 @@ class Neo4jWriteMappingStrategy(private val options: Neo4jOptions)
     if (nodeMetadata.nodeKeys.contains(key)) {
       nodeMap.get(KEYS).put(nodeMetadata.nodeKeys.getOrElse(key, key), value)
     }
-    if (nodeMetadata.nodeProps.contains(key)) {
-      nodeMap.get(PROPERTIES).put(nodeMetadata.nodeProps.getOrElse(key, key), value)
+    if (nodeMetadata.properties.contains(key)) {
```
fbiville, line 76: missing `else`?

conker84: No, it's not if/else alternative we need to check for both `nodeKeys` and `properties`, I guess that we can change the method in this way:

```scala
    if (nodeMetadata.nodeKeys.contains(key)) {
      nodeMap.get(KEYS).put(nodeMetadata.nodeKeys.get(key), value)
    }
    if (nodeMetadata.properties.contains(key)) {
      nodeMap.get(PROPERTIES).put(nodeMetadata.properties.get(key), value)
    }
```

fbiville (approved):

fbiville (changes requested):
We discussed this a bit further internally and would like to simplify the configuration surface, details to follow soon.

common/src/main/scala/org/neo4j/spark/service/SchemaService.scala
```diff
@@ -628,9 +628,9 @@ class SchemaService(private val options: Neo4jOptions, private val driverCache:
 
   private def createEntityConstraint(entityType: String,
                                      entityIdentifier: String,
-                                     unique: Boolean,
+                                     constraintsOptimizationType: ConstraintsOptimizationType.Value,
                                      keys: Map[String, String]): Unit = {
-    val constraintType = if (unique) {
+    val constraintType = if (constraintsOptimizationType == ConstraintsOptimizationType.UNIQUE) {
       "UNIQUE"
     } else {
```
fbiville, line 635: what about `NONE`?

common/src/main/scala/org/neo4j/spark/util/Neo4jOptions.scala
```diff
@@ -435,11 +437,9 @@ object Neo4jOptions {
   // deprecated in favor of...
   val SCHEMA_OPTIMIZATION_TYPE = "schema.optimization.type"
   // ...these options
-  val SCHEMA_OPTIMIZATION_NODE_UNIQUE = "schema.optimization.node.unique"
+  val SCHEMA_OPTIMIZATION = "schema.optimization"
   val SCHEMA_OPTIMIZATION_NODE_KEY = "schema.optimization.node.key"
```
fbiville: should be `.keys`

common/src/main/scala/org/neo4j/spark/util/Neo4jOptions.scala
```diff
@@ -435,11 +437,9 @@ object Neo4jOptions {
...
-  val SCHEMA_OPTIMIZATION_RELATIONSHIP_UNIQUE = "schema.optimization.relationship.unique"
   val SCHEMA_OPTIMIZATION_RELATIONSHIP_KEY = "schema.optimization.relationship.key"
```
fbiville: should be `.keys`

fbiville (changes requested):

common/src/main/scala/org/neo4j/spark/service/SchemaService.scala
fbiville: I guess `createEntityConstraint` is not called when the optimization is `NONE`?

common/src/main/scala/org/neo4j/spark/util/Neo4jOptions.scala
conker84: done

common/src/main/scala/org/neo4j/spark/util/Neo4jOptions.scala
conker84: done

fbiville (approved):

PR #583, opened 2023-12-20
Improved user-agent management

conker84:
Added user agent information about Databricks if available

### DIFF
--- a/common/src/main/scala/org/neo4j/spark/util/Neo4jOptions.scala
+++ b/common/src/main/scala/org/neo4j/spark/util/Neo4jOptions.scala
@@ -295,7 +295,7 @@ case class Neo4jDriverOptions(
 
   def toDriverConfig: Config = {
     val builder = Config.builder()
-      .withUserAgent(s"neo4j-spark-connector/${Neo4jUtil.connectorVersion}")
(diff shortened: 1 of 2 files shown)


common/src/main/scala/org/neo4j/spark/util/Neo4jOptions.scala
```diff
@@ -295,7 +295,7 @@ case class Neo4jDriverOptions(
 
   def toDriverConfig: Config = {
     val builder = Config.builder()
-      .withUserAgent(s"neo4j-spark-connector/${Neo4jUtil.connectorVersion}")
+      .withUserAgent(s"neo4j-spark-connector/${Neo4jUtil.connectorVersion}/${Neo4jUtil.connectorEnv}")
```
fbiville: maybe avoid using `/` as a separator between connectionVersion and the databricks info, as this goes against the recommendation in https://neo4j.com/docs/bolt/current/bolt/message/#messages-hello
```suggestion
      .withUserAgent(s"neo4j-spark-connector/${Neo4jUtil.connectorVersion} (${Neo4jUtil.connectorEnv})")
```

fbiville (changes requested):

common/src/main/scala/org/neo4j/spark/util/Neo4jOptions.scala
conker84: Done

fbiville (approved):

PR #584, opened 2023-12-20
Improved query error management

conker84:
Improved query error management now, if a query has an error we don't return only `Please provide a valid <TYPE> query` but a more elaborate message like `Query count not compiled for the following exception: <error from the compiler>`

### DIFF
--- a/common/src/main/scala/org/neo4j/spark/service/SchemaService.scala
+++ b/common/src/main/scala/org/neo4j/spark/service/SchemaService.scala
@@ -1,5 +1,6 @@
 package org.neo4j.spark.service
 
+import org.apache.commons.lang3.exception.ExceptionUtils
 import org.apache.spark.internal.Logging
 import org.apache.spark.sql.sources.Filter
 import org.apache.spark.sql.types.{DataType, DataTypes, StructField, StructType}
@@ -509,6 +510,41 @@ class SchemaService(private val options: Neo4jOptions, private val driverCache: 
       .asBoolean()
   }
 
+  def validateQuery(query: String, expectedQueryTypes: org.neo4j.driver.summary.QueryType*): String = try {
+    val queryType = session.run(s"EXPLAIN $query").consume().queryType()
+    if (expectedQueryTypes.isEmpty || expectedQueryTypes.contains(queryType)) {
+      ""
+    } else {
+      s"Invalid query `${cleanQuery(query)}` because the accepted types are [${expectedQueryTypes.mkString(", ")}], but the actual type is $queryType"
+    }
+  } catch {
+    case e: Throwable => s"Query not compiled for the following exception: ${ExceptionUtils.getMessage(e)}"
+  }
+
+  private def cleanQuery(query: String) = {
+    query
+      .replace(s"WITH {} AS ${Neo4jQueryStrategy.VARIABLE_EVENT}, [] as ${Neo4jQueryStrategy.VARIABLE_SCRIPT_RESULT}", "")
+      .replace(s"WITH [] as ${Neo4jQueryStrategy.VARIABLE_SCRIPT_RESULT}", "")
+      .replace(s"WITH {} AS ${Neo4jQueryStrategy.VARIABLE_EVENT}", "")
+      .trim
+  }
+
+  def validateQueryCount(query: String): String = try {
+    val resultSummary = session.run(s"EXPLAIN $query").consume()
+    val queryType = resultSummary.queryType()
+    val plan = resultSummary.plan()
+    val expectedQueryTypes = Set(org.neo4j.driver.summary.QueryType.READ_ONLY, org.neo4j.driver.summary.QueryType.SCHEMA_WRITE)
+    val isReadOnly = expectedQueryTypes.contains(queryType)
+    val hasCountIdentifier = plan.identifiers().asScala.toSet == Set("count")
+    if (isReadOnly && hasCountIdentifier) {
+      ""
+    } else {
+      s"Invalid query `${cleanQuery(query)}` because the expected type should be [${expectedQueryTypes.mkString(", ")}], but the actual type is $queryType"
+    }
+  } catch {
+    case e: Throwable => s"Query count not compiled for the following exception: ${ExceptionUtils.getMessage(e)}"
+  }
+
   def isValidQuery(query: String, expectedQueryTypes: org.neo4j.driver.summary.QueryType*): Boolean = try {
     val queryType = session.run(s"EXPLAIN $query").consume().queryType()
     expectedQueryTypes.isEmpty || expectedQueryTypes.contains(queryType)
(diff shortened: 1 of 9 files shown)


common/src/main/scala/org/neo4j/spark/service/SchemaService.scala
```diff
@@ -509,6 +510,33 @@ class SchemaService(private val options: Neo4jOptions, private val driverCache:
       .asBoolean()
   }
 
+  def validateQuery(query: String, expectedQueryTypes: org.neo4j.driver.summary.QueryType*): String = try {
+    val queryType = session.run(s"EXPLAIN $query").consume().queryType()
+    if (expectedQueryTypes.isEmpty || expectedQueryTypes.contains(queryType)) {
+      ""
+    } else {
+      s"Query non valid because the expected type should be [${expectedQueryTypes.mkString(", ")}], but the actual type is $queryType"
```
fbiville: what about including the query as well in the error message?
```suggestion
      s"Invalid query $query because the accepted types are [${expectedQueryTypes.mkString(", ")}], but the actual type is $queryType"
```

common/src/main/scala/org/neo4j/spark/service/SchemaService.scala
```diff
@@ -509,6 +510,33 @@ class SchemaService(private val options: Neo4jOptions, private val driverCache:
...
+    }
+  } catch {
+    case e: Throwable => s"Query not compiled for the following exception: ${ExceptionUtils.getMessage(e)}"
```
fbiville, line 521: same question as above

common/src/main/scala/org/neo4j/spark/service/SchemaService.scala
```diff
@@ -509,6 +510,33 @@ class SchemaService(private val options: Neo4jOptions, private val driverCache:
...
+    val resultSummary = session.run(s"EXPLAIN $query").consume()
+    val queryType = resultSummary.queryType()
+    val plan = resultSummary.plan()
+    val expectedQueryTypes = Set(org.neo4j.driver.summary.QueryType.READ_ONLY, org.neo4j.driver.summary.QueryType.SCHEMA_WRITE)
+    val isReadOnly = expectedQueryTypes.contains(queryType)
+    val hasCountIdentifier = plan.identifiers().asScala.toSet == Set("count")
+    if (isReadOnly && hasCountIdentifier) {
+      ""
+    } else {
+      s"Query count non valid because the expected type should be [${expectedQueryTypes.mkString(", ")}], but the actual type is $queryType"
```
fbiville: same question as above

common/src/main/scala/org/neo4j/spark/service/SchemaService.scala
```diff
@@ -509,6 +510,33 @@ class SchemaService(private val options: Neo4jOptions, private val driverCache:
...
+    }
+  } catch {
+    case e: Throwable => s"Query count not compiled for the following exception: ${ExceptionUtils.getMessage(e)}"
```
fbiville, line 545: same question as above

common/src/test/scala/org/neo4j/spark/util/ValidationsIT.scala
```diff
@@ -0,0 +1,180 @@
+package org.neo4j.spark.util
+
+import org.junit.rules.ExpectedException
+import org.junit.{Rule, Test}
+import org.neo4j.driver.AccessMode
+import org.neo4j.spark.{SparkConnectorScalaSuiteIT, TestUtil}
+
+class ValidationsIT extends SparkConnectorScalaSuiteIT {
+
+  val _expectedException: ExpectedException = ExpectedException.none
```
fbiville: why is the extra field needed?

common/src/test/scala/org/neo4j/spark/util/ValidationsIT.scala
```diff
@@ -0,0 +1,180 @@
...
+
+  @Rule
+  def exceptionRule: ExpectedException = _expectedException
+
+  @Test
+  def testReadQueryShouldBeSyntacticallyInvalid(): Unit = {
+    // then
+    val expectedMessage = if (TestUtil.neo4jVersion().startsWith("4")) {
```
fbiville: this assertion may be a bit too strict and may unfortunately change again in future versions, which will require updates to this test
maybe just check for "Invalid input"?

common/src/test/scala/org/neo4j/spark/util/ValidationsIT.scala
```diff
@@ -0,0 +1,180 @@
...
+    readOpts.put("query", "MERGE (n:TestNode{id: 1}) RETURN n")
+
+    // when
+    Validations.validate(ValidateRead(new Neo4jOptions(readOpts), "1"))
+  }
+
+  @Test
+  def testReadQueryCountBeSyntacticallyInvalid(): Unit = {
+    // then
+    val expectedMessage = if (TestUtil.neo4jVersion().startsWith("4")) {
```
fbiville: same as above

common/src/main/scala/org/neo4j/spark/service/SchemaService.scala
conker84: Oki

common/src/main/scala/org/neo4j/spark/service/SchemaService.scala
conker84: In this case the query cames from the compiler itself so it's already included. Please check the tests

common/src/main/scala/org/neo4j/spark/service/SchemaService.scala
conker84: In this case the query cames from the compiler itself so it's already included. Please check the tests

common/src/main/scala/org/neo4j/spark/service/SchemaService.scala
conker84: In this case the query cames from the compiler itself so it's already included. Please check the tests

common/src/test/scala/org/neo4j/spark/util/ValidationsIT.scala
conker84: It's a workaround otherwise rules does not work. Honestly I don't remember the reason why; please check this discussion (https://stackoverflow.com/questions/32160549/using-junit-rule-with-scalatest-e-g-temporaryfolder)

fbiville: https://stackoverflow.com/a/47068754/277128 seems slightly better?

common/src/test/scala/org/neo4j/spark/util/ValidationsIT.scala
conker84: Done

common/src/main/scala/org/neo4j/spark/service/SchemaService.scala
fbiville: I'm on the fence, since we do not really own the error message and the query could be removed there, but I guess we can keep it like this for now.

common/src/test/scala/org/neo4j/spark/util/ValidationsIT.scala
fbiville:
```suggestion
  val expectedException: ExpectedException = ExpectedException.none
```

fbiville (approved):

PR #614, opened 2024-04-22
build: add package for spark-packages

venikkin:
* add assembly configs for spark-packages
* add a script to upload the file


pom.xml
```diff
@@ -208,25 +208,6 @@
                     </jvmArgs>
                 </configuration>
             </plugin>
-            <plugin>
```
fbiville: doesn't this deletion break the regular deployment to Maven Central?

fbiville (commented):
1 question only, the rest LGTM

pom.xml
venikkin: This is a good question. For the last 2 releases as far as I can see we are releasing only spark 3. Let me double-check

maven-release.sh
```diff
@@ -65,6 +65,13 @@ sed_i "s/<artifactId>neo4j-connector-apache-spark_parent<\/artifactId>/<artifact
 sed_i "s/<artifactId>neo4j-connector-apache-spark_common<\/artifactId>/<artifactId>neo4j-connector-apache-spark_${SCALA_VERSION}_common<\/artifactId>/" "spark-3/pom.xml"
 sed_i "s/<artifactId>neo4j-connector-apache-spark_test-support<\/artifactId>/<artifactId>neo4j-connector-apache-spark_${SCALA_VERSION}_test-support<\/artifactId>/" "spark-3/pom.xml"
 
+LATEST_SCALA_VERSION="2.13"
```
fbiville: so Spark packages assume the latest-at-the-deploy-time Scala version by default?

Let's take an example:

- At t1, Scala latest is Scala 2.13 and Spark Connector `5.4.3` is released (without suffix)
- At t2>t1, Scala latest is now Scala 3. The `5.4.3` release is still for Scala 2.13, hopefully, right? If I release 5.4.4 now, it would be assumed that the release targets Scala 3?

We could always suffix if we wanted to, or?

venikkin: You are right. If we support multiple scala versions, we need to prefix all versions. Will change it.

pom.xml
venikkin: I didn't expect it, but we do release all submodules as jars with deps. Even `test-support` (https://mvnrepository.com/artifact/org.neo4j/neo4j-connector-apache-spark_2.12_test-support/5.3.0 - 33Mb). I'm not sure why we need to publish them, but it's safer to keep it as it is and change it globally.

fbiville (approved):

venikkin:
[Investigate publication of Spark packages to Databricks](https://trello.com/c/FdHE70Zu/457-investigate-publication-of-spark-packages-to-databricks)

PR #625, opened 2024-06-05
Remove Claudio from code owners

fbiville:


venikkin (approved):
We probably need to add @Emrehzl94 to the `.github` folder owners

PR #626, opened 2024-06-06
Fix how resolvers are handled during driver construction

ali-ince:
This PR fixes how multiple URLs are handled when passed to `url` option during driver creation.

### DIFF
--- a/common/src/main/scala/org/neo4j/spark/util/Neo4jOptions.scala
+++ b/common/src/main/scala/org/neo4j/spark/util/Neo4jOptions.scala
@@ -15,6 +15,7 @@ import java.util.concurrent.TimeUnit
 import scala.collection.JavaConverters._
 import scala.language.implicitConversions
 import org.apache.spark.internal.Logging
+import org.jetbrains.annotations.TestOnly
 
 
 class Neo4jOptions(private val options: java.util.Map[String, String]) extends Serializable with Logging {
@@ -329,7 +330,12 @@ case class Neo4jDriverOptions(
                                connectionTimeout: Int
                              ) extends Serializable {
 
-  def toDriverConfig: Config = {
+  def createDriver(): Driver = {
+    val (url, _) = connectionUrls
+    GraphDatabase.driver(url, toNeo4jAuth, toDriverConfig)
+  }
+
+  private def toDriverConfig: Config = {
     val builder = Config.builder()
       .withUserAgent(s"neo4j-${Neo4jUtil.connectorEnv}-connector/${Neo4jUtil.connectorVersion}")
       .withLogging(Logging.slf4j())
@@ -370,18 +376,20 @@ case class Neo4jDriverOptions(
   }
 
   // public only for testing purposes
+  @TestOnly
   def connectionUrls: (URI, Set[ServerAddress]) = {
     val urls = url.split(",").toList
-    val extraUrls = urls
+    val resolved = urls
       .drop(1)
       .map(_.trim)
       .map(URI.create)
       .map(uri => ServerAddress.of(uri.getHost, if (uri.getPort > -1) uri.getPort else 7687))
       .toSet
-    (URI.create(urls.head.trim), extraUrls)
+    (URI.create(urls.head.trim), resolved)
   }
 
   // public only for testing purposes
+  @TestOnly
   def toNeo4jAuth: AuthToken = {
     auth match {
       case "basic" => AuthTokens.basic(username, password)
--- a/common/src/main/scala/org/neo4j/spark/util/Neo4jOptions.scala
+++ b/common/src/main/scala/org/neo4j/spark/util/Neo4jOptions.scala
@@ -31,7 +31,7 @@ class Neo4jOptions(private val options: java.util.Map[String, String]) extends S
         _.conf
           .getAll
           .filterKeys(k => k.startsWith("neo4j."))
-          .map { elem => (elem._1.substring("neo4.".length + 1), elem._2) }
+          .map { elem => (elem._1.substring("neo4j.".length), elem._2) }
           .toMap
       }
       .getOrElse(Map.empty)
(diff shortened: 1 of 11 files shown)


common/src/main/scala/org/neo4j/spark/util/Neo4jOptions.scala
```diff
@@ -15,21 +15,24 @@ import java.util.concurrent.TimeUnit
 import scala.collection.JavaConverters._
 import scala.language.implicitConversions
 import org.apache.spark.internal.Logging
+import org.jetbrains.annotations.TestOnly
```
fbiville, line 18: this addition is probably not intentional?

common/src/main/scala/org/neo4j/spark/util/Neo4jOptions.scala
```diff
@@ -15,21 +15,24 @@ import java.util.concurrent.TimeUnit
...
-      .map { _.conf
-        .getAll
-        .filterKeys(k => k.startsWith("neo4j."))
-        .map { elem => (elem._1.substring("neo4.".length + 1), elem._2) }
-        .toMap
+      .map {
+        _.conf
+          .getAll
+          .filterKeys(k => k.startsWith("neo4j."))
+          .map { elem => (elem._1.substring("neo4.".length + 1), elem._2) }
```
fbiville: why "neo4" and not "neo4j"?

fbiville: nvm, t'was already there

common/src/test/scala/org/neo4j/spark/util/Neo4jOptionsIT.scala
```diff
@@ -0,0 +1,50 @@
...
+      use(driver.session()) { session =>
+        assertEquals(1, session.run("RETURN 1").single().get(0).asInt())
+      }
+    }
+  }
+
+  @Test
+  def shouldConstructDriverWithResolver(): Unit = {
+    val options: java.util.Map[String, String] = new java.util.HashMap[String, String]()
+    options.put(Neo4jOptions.URL, s"neo4j://localhost.localdomain:8888, bolt://localhost.localdomain:9999, ${server.getBoltUrl}")
```
fbiville, line 31: wait, how does this work?

common/src/main/scala/org/neo4j/spark/util/Neo4jOptions.scala
```diff
@@ -329,7 +336,12 @@ case class Neo4jDriverOptions(
                                connectionTimeout: Int
                              ) extends Serializable {
 
-  def toDriverConfig: Config = {
+  def createDriver(): Driver = {
+    val (url, _) = connectionUrls
```
fbiville, line 340: does it ignore the other URLs? Shouldn't we use a resolver to fall back on these extra URLs instead?

common/src/main/scala/org/neo4j/spark/util/Neo4jOptions.scala
```diff
@@ -370,18 +382,20 @@ case class Neo4jDriverOptions(
   }
 
   // public only for testing purposes
+  @TestOnly
   def connectionUrls: (URI, Set[ServerAddress]) = {
     val urls = url.split(",").toList
-    val extraUrls = urls
+    val resolved = urls
       .drop(1)
```
fbiville, line 389: nit: you could move the `drop(1)` to the end, so that all URLs are processed in the same way

common/src/main/scala/org/neo4j/spark/util/Neo4jOptions.scala
```diff
@@ -370,18 +382,20 @@ case class Neo4jDriverOptions(
...
       .map(_.trim)
       .map(URI.create)
       .map(uri => ServerAddress.of(uri.getHost, if (uri.getPort > -1) uri.getPort else 7687))
```
fbiville, line 392: nit: I think the port handling is actually a job better left done by the driver

fbiville (approved):

common/src/main/scala/org/neo4j/spark/util/Neo4jOptions.scala
ali-ince: I actually added it two public functions on this class, as they were already commented as being public for test access.

common/src/main/scala/org/neo4j/spark/util/Neo4jOptions.scala
ali-ince: Fixed this

common/src/test/scala/org/neo4j/spark/util/Neo4jOptionsIT.scala
ali-ince: So, I was just mapping `localhost.localdomain:8888` to a set of `localhost.localdomain:9999` (which is not accessible) and `localhost:XYZ` (which should be accessible) - and I expect the driver to establish a connection successfully.

Bu we need a response from drivers team for this - I believe there is a bug around this in the driver.

common/src/main/scala/org/neo4j/spark/util/Neo4jOptions.scala
ali-ince: resolver function is being built as part of `toDriverConfig`, this is just the first entry of the `url` parameter returned as a `java.net.URI` object.

common/src/main/scala/org/neo4j/spark/util/Neo4jOptions.scala
ali-ince: yep, tried it but this `resolved` is being converted into `ServerAddress` set which is not `URI` - so I'll keep it as it is for now.

common/src/main/scala/org/neo4j/spark/util/Neo4jOptions.scala
ali-ince: I think we need to handle this here, as we are constructing an explicit `ServerAddress` with an explicit hostname and a port number which is validated by `BoltServerAddress`.

ali-ince:
waiting for a driver _fix_ for this one.

fbiville (approved):

PR #660, opened 2024-09-08
feat: support checkpointing on structured streams

### LINKED ISSUE #579: Streaming reads: offsets are not loaded from checkpoint when restarting a stream after failure.
## Expected Behavior (Mandatory)

When restarting a stream after failure or termination with the same checkpointLocation, the stream should restart from the latest offset.

## Actual Behavior (Mandatory)

The current timestamp (`"streaming.from", "NOW"`) or 0 (`"streaming.from", "ALL"`) is used instead.

## How to Reproduce the Problem

Create a read stream with `"streaming.from", "ALL"`. Start the query, stop the query, then start the query again. The query restarts with offset 0. 

### Simple Dataset (where it's possible)

```
// Insert the output of the `df.show()` call

```

### Steps (Mandatory)

1.
1.
1.

## Specifications (Mandatory)

Currently used versions

### Versions

- Spark: 3.4.0
- Scala: 2.12
- Neo4j: 5.10.0
- Neo4j Connector: 5.2.0

## Additional information

Disclaimer: I am not an expert. If I'm just doing something wrong, please let me know. :)

The neo4j connector appears to be using an Accumulator to store offsets:

https://github.com/neo4j-contrib/neo4j-spark-connector/blob/4071559212bc4a898aa9ca9f5ef8324c328d3293/common/src/main/scala/org/neo4j/spark/streaming/OffsetStorage.scala#L28

Accumulators are not checkpointed so don't survive a restart. This means that on restarting the stream, the initial offset is reported as the latest offset:

https://github.com/neo4j-contrib/neo4j-spark-connector/blob/4071559212bc4a898aa9ca9f5ef8324c328d3293/spark-3/src/main/scala/org/neo4j/spark/streaming/Neo4jMicroBatchReader.scala#L75

Does the offset need to be written to the checkpoint on commit and read on stream restart? I have seen this pattern in other connectors (e.g. https://github.com/mongodb/mongo-spark/blob/main/src/main/java/com/mongodb/spark/sql/connector/read/MongoOffsetStore.java.)

ali-ince:
This PR updates structured streaming parts of the connector so that it relies to offsets checkpointed rather than following it's own logic.

Fixes #579.


spark-3/src/test/scala/org/neo4j/spark/DataSourceStreamingReaderTSE.scala
```diff
@@ -44,352 +48,301 @@ class DataSourceStreamingReaderTSE extends SparkConnectorScalaBaseTSE {
...
-          Thread.sleep(200)
-          SparkConnectorScalaSuiteIT.session()
-            .writeTransaction(new TransactionWork[ResultSummary] {
-              override def execute(tx: Transaction): ResultSummary = {
-                tx.run(s"CREATE (n:Test4_Movie {title: 'My movie $index', timestamp: timestamp()})")
-                  .consume()
-              }
-            })
-        })
+        createMovieNodes(1, 60, 1000, 200)
```
venikkin:
```suggestion
        createMovieNodes(1, total, 1000, 200)
```

venikkin (approved):

PR #675, opened 2024-10-17
ci: add python 3.13 to test matrix

venikkin:


.github/workflows/python-ci.yml
```diff
@@ -13,11 +13,14 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [ "3.8", "3.9", "3.10", "3.11", "3.12" ]
+        python-version: [ "3.8", "3.9", "3.10", "3.11", "3.12", "3.13" ]
```
ali-ince, line 16: `3.8` is also EOL'd earlier this month, should we remove that too?

https://devguide.python.org/versions/

.github/workflows/python-ci.yml
```diff
@@ -13,11 +13,14 @@ jobs:
...
         neo4j-version: [ "4.4", "5" ]
         spark-version:
           - { short: "3", ext: "3.4.3", scala: "2.12" }
           - { short: "3", ext: "3.5.2", scala: "2.12" }
+        exclude:
+          - python-version: "3.13"
+          - spark-version: { short: "3", ext: "3.4.3", scala: "2.12" }
```
ali-ince: Does this not exclude all python 3.13 builds?

Should it look like this instead?

```suggestion
        exclude:
          - python-version: "3.13"
            spark-version: { short: "3", ext: "3.4.3", scala: "2.12" }
```

fbiville: argh 🤦 let me fix this

.github/workflows/python-ci.yml
fbiville: follow-up PR OK for this?

fbiville: https://github.com/neo4j/neo4j-spark-connector/pull/678

ali-ince (approved):

PR #692, opened 2025-01-24
feat: support version 2025

venikkin:
Add basic support for newcoming neo4j 2025.

### DIFF
--- a/common/src/main/scala/org/neo4j/spark/service/SchemaService.scala
+++ b/common/src/main/scala/org/neo4j/spark/service/SchemaService.scala
@@ -666,20 +666,19 @@ class SchemaService(
           val quotedProps = props
             .map(prop => s"${Neo4jUtil.NODE_ALIAS}.${prop.quote()}")
             .mkString(", ")
-          val isNeo4j5 =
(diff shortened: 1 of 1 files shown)


common/src/main/scala/org/neo4j/spark/service/SchemaService.scala
```diff
@@ -666,7 +666,8 @@ class SchemaService(
           val quotedProps = props
             .map(prop => s"${Neo4jUtil.NODE_ALIAS}.${prop.quote()}")
             .mkString(", ")
-          val isNeo4j5 = neo4jVersion().versions(0).startsWith("5.")
+          val isNeo4j5 =
```
ali-ince: should we just revert this check instead? like if it is "4." then do the special logic but assume later versions instead?

ali-ince (changes requested):

common/src/main/scala/org/neo4j/spark/service/SchemaService.scala
fbiville: fixed now

ali-ince (approved):

PR #698, opened 2025-02-14
feat: prepend Cypher version when supported

fbiville:


spark-3/src/main/scala/org/neo4j/spark/DataSource.scala
```diff
@@ -38,23 +42,42 @@ class DataSource extends TableProvider
...
+        )
     }
 
     schema
   }
 
+  private def getNeo4jInfo(options: Neo4jDriverOptions): Neo4j = {
+    if (neo4j == null) {
+      val driver = new DriverCache(options).getOrCreate()
+      neo4j = Neo4jDetectorKt.detectedWith(Neo4j.Companion, driver)
```
ali-ince: Have you tried `@JvmStatic` and `@JvmName` annotations on Kotlin side, so that we get rid of `Kt` or `Companion` uses?

common/src/test/scala/org/neo4j/spark/service/Neo4jQueryServiceTest.scala
```diff
@@ -38,14 +42,17 @@ import scala.collection.immutable.HashMap
 
 class Neo4jQueryServiceTest {
 
+  private val neo4j: Neo4j =
+    new Neo4j(new Neo4jVersion(4, 4, 0), Neo4jEdition.ENTERPRISE, Neo4jDeploymentType.SELF_MANAGED)
```
ali-ince: Do we also need to add some tests for later versions that end up in `CYPHER 5` prefix?

ali-ince (commented):
Left two comments, otherwise looks good to me.

spark-3/src/main/scala/org/neo4j/spark/DataSource.scala
fbiville: If I do that, it will slightly degrade the experience in Kotlin, since `detectedWith` is defined as an extension function and called like this: `Neo4j.detectedWith(driver)`.

The benefit of the existing approach is that `Neo4j` remains the sole entry point.

I can of course deprecate it and introduce `Neo4jDetector.detect(driver)` as a replacement, but it's slightly less discoverable (but easier to use outside Kotlin).

common/src/test/scala/org/neo4j/spark/service/Neo4jQueryServiceTest.scala
fbiville: added in 069e9bf99d65e506c03240d67ce6bbea6532bdf7

spark-3/src/main/scala/org/neo4j/spark/DataSource.scala
fbiville: Proposal is here: https://github.com/neo4j/caniuse/pull/18

ali-ince (approved):

PR #708, opened 2025-03-04
feat: add support for neo4j driver transaction timeout

dhrudevalia:
## Feature
- Expose `neo4j.db.transaction.timeout` to Spark connector configuration
- Update schema service writeTransaction to include timeout if set
- Update partition and data writers readTransaction to include timeout if set

### DIFF
--- a/common/src/main/scala/org/neo4j/spark/util/Neo4jOptions.scala
+++ b/common/src/main/scala/org/neo4j/spark/util/Neo4jOptions.scala
@@ -325,6 +325,19 @@ class Neo4jOptions(private val options: java.util.Map[String, String]) extends S
     StreamingFrom.withCaseInsensitiveName(getParameter(STREAMING_FROM, DEFAULT_STREAMING_FROM.toString)),
     getParameter(STREAMING_QUERY_OFFSET)
   )
+
+  def toNeo4jTransactionConfig(): TransactionConfig = {
+    val timeout = getParameter(TRANSACTION_TIMEOUT_MSECS, DEFAULT_TRANSACTION_TIMEOUT_MSECS.toString).toInt
+
+    val builder = TransactionConfig.builder()
+    if (timeout > 0) {
+      val duration = Duration.ofMillis(timeout)
+      builder.withTimeout(duration)
+    }
+
+    builder.build()
+  }
+
 }
 
 case class Neo4jStreamingOptions(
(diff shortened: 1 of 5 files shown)


common/src/main/scala/org/neo4j/spark/util/Neo4jOptions.scala
```diff
@@ -325,6 +325,19 @@ class Neo4jOptions(private val options: java.util.Map[String, String]) extends S
     StreamingFrom.withCaseInsensitiveName(getParameter(STREAMING_FROM, DEFAULT_STREAMING_FROM.toString)),
     getParameter(STREAMING_QUERY_OFFSET)
   )
+
+  def toNeo4jTransactionConfig(): TransactionConfig = {
+    val timeout = getParameter(TRANSACTION_TIMEOUT_MSECS, DEFAULT_TRANSACTION_TIMEOUT_MSECS.toString).toInt
+
+    val builder = TransactionConfig.builder()
+    if (timeout > 0) {
```
venikkin: Maybe we can replace this with null-checks. According to [this](https://neo4j.com/docs/api/java-driver/current/org.neo4j.driver/org/neo4j/driver/TransactionConfig.Builder.html#withTimeout(java.time.Duration)) `0` means indefinite transaction time. When the default is the server-side default. Filtering out `0` will make indefinite timeouts not possible.

venikkin (changes requested):

common/src/main/scala/org/neo4j/spark/util/Neo4jOptions.scala
dhrudevalia: The TransactionConfig doesn't allow 0 values unfortunately.

From `withTimeout()`
```
public Builder withTimeout(Duration timeout) {
    requireNonNull(timeout, "Transaction timeout should not be null");
    checkArgument(!timeout.isZero(), "Transaction timeout should not be zero");
    checkArgument(!timeout.isNegative(), "Transaction timeout should not be negative");
...
```

So I thought for an intentional fallback it would be better to be 0 since it has to be the same type. Then we do a nonzero check and set it for the TransactionConfig.

venikkin: You are right. Looks like 0 was added in 5.x version. 
```java
        public Builder withTimeout(Duration timeout) {
            Objects.requireNonNull(timeout, "Transaction timeout should not be null");
            Preconditions.checkArgument(!timeout.isNegative(), "Transaction timeout should not be negative");
            this.timeout = timeout;
            return this;
        }
```
I think it fine, but maybe we still can go with the null change to make this feature-proof to not forget about this when we migrate to 5.x?

dhrudevalia: So just to confirm the change would be to set default of transaction timeout to null, and then use the same check but instead of checking for `>0` check for null.

dhrudevalia: Updated in latest commit.

venikkin (approved):
Looks good

PR #726, opened 2025-04-22
build: default to Scala 2.12

fbiville:
It is much more unlikely to face issues when compiling Scala 2.12 code with 2.13 than the other way around.

Note that CI and deployment script will continue to work as they always pass `-Dscala-${scala_version}` to Maven.


venikkin (approved):
Snyk has complained about a few vulnerabilities in Scala 2.12, but I think we need to deal with them separately, as we are not introducing new dependencies in this PR. 

Looks good to me.

PR #728, opened 2025-05-06
ci: setup teamcity based build pipeline

ali-ince:
This PR sets up Teamcity based build pipeline, including daily compatibility builds and a release build.


.teamcity/builds/Build.kt
```diff
@@ -0,0 +1,153 @@
...
+          val bts = sequential {
+            if (forPullRequests)
+                buildType(WhiteListCheck("${name}-whitelist-check", "white-list check"))
+            if (forPullRequests) dependentBuildType(PRCheck("${name}-pr-check", "pr check"))
+
+            parallel {
+              javaVersions.cartesianProduct(scalaVersions, neo4jVersions).forEach {
+                  (java, scala, neo4j) ->
+                sequential {
+                  val packaging =
```
venikkin, line 35: I wonder if it makes sense to re-use this package step based only on `java` and `scala` version for all dependent test builds for each `neo4j` version?  

Maybe we can have a nested loop over neo4j version similarly as we have pythin version nested loop?

venikkin (approved):
Looks good. One comment about build re-use

.teamcity/builds/Build.kt
ali-ince: That was my initial attempt too, but Teamcity complains if a single build is re-used by different contexts and I thought this would be the most straightforward way to make it work (and since we have caching in place it wouldn't be a big loss of time/resources).

PR #730, opened 2025-05-06
build: do not use an URL shortener for downloading JReleaser

michael-simons:
Feels a bit wrong, better use at least a tagged Github reference.


fbiville (approved):

ali-ince (approved):
This definitely looks better @michael-simons, thanks!

PR #738, opened 2025-05-16
feat: add retries for reader queries

dhrudevalia:
## Issue

Retry logic is currently only implemented for write queries.

## Fix
Implement retry logic for reads in `BasePartitionReader.scala`. The same retry amount configuration option is used.

### DIFF
--- a/common/src/main/scala/org/neo4j/spark/reader/BasePartitionReader.scala
+++ b/common/src/main/scala/org/neo4j/spark/reader/BasePartitionReader.scala
@@ -38,10 +38,10 @@ import org.neo4j.spark.util.Neo4jUtil
 import org.neo4j.spark.util.QueryType
 
 import java.io.IOException
+import java.time.Duration
 import java.util
 import java.util.concurrent.CountDownLatch
 import java.util.concurrent.locks.LockSupport
-import java.time.Duration
 
 import scala.collection.JavaConverters._
 
--- a/common/src/main/scala/org/neo4j/spark/reader/BasePartitionReader.scala
+++ b/common/src/main/scala/org/neo4j/spark/reader/BasePartitionReader.scala
@@ -39,6 +39,9 @@ import org.neo4j.spark.util.QueryType
 
 import java.io.IOException
 import java.util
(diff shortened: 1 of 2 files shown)


common/src/main/scala/org/neo4j/spark/reader/BasePartitionReader.scala
```diff
@@ -81,37 +84,63 @@ abstract class BasePartitionReader(
   @volatile
   private var error: Boolean = false
 
+  private val retries = new CountDownLatch(options.transactionSettings.retries)
```
venikkin: Usually, `CountDownLatch` is used because it can block the thread until the countdown is complete. You don't use this semantics here, so the regular `AtomicInteger` or even regular `short` counter (not sure about the later as don't know how concurrent this class is and how atomic this counter should be) should suffice.

venikkin (changes requested):

common/src/main/scala/org/neo4j/spark/reader/BasePartitionReader.scala
dhrudevalia: Sounds good. Will update this. I was just replicating the retry logic for the writer class.

venikkin (approved):

common/src/main/scala/org/neo4j/spark/reader/BasePartitionReader.scala
```diff
@@ -81,37 +84,63 @@ abstract class BasePartitionReader(
...
-      }
+      case t: Throwable =>
+        if (options.transactionSettings.shouldFailOn(t)) {
+          error = true
+          logError("Error while invoking next due to explicitly configured failure condition:", t)
+          throw new IOException(t)
+        }
+
+        if (Neo4jUtil.isRetryableException(t) && retries.get() > 0) {
+          val currentRetry = retries.decrementAndGet
```
venikkin, line 102: This is not very atomic operation and it can increase the number of retries if used in multiple threads, but I don't think it's a problem.

PR #754, opened 2025-08-12
feat: add support for authentication token providers

venikkin:
Added support for token supplier. All existing authentication methods are replaced with provided from `commons-authn-provided` library. Additionally, it is possible to add extra token provider via service loader.


common/src/main/scala/org/neo4j/spark/auth/BasicAuthenticationSupplierFactory.scala
```diff
@@ -0,0 +1,40 @@
...
+package org.neo4j.spark.auth
+
+import org.neo4j.connectors.authn.AuthenticationToken
+import org.neo4j.connectors.authn.AuthenticationTokenSupplierFactory
+import org.neo4j.spark.util.Neo4jOptions.DEFAULT_EMPTY
+
+import java.util
+import java.util.function.Supplier
+
+class BasicAuthenticationSupplierFactory extends AuthenticationTokenSupplierFactory {
```
ali-ince: How about moving these factories into connectors-commons?

fbiville (approved):

PR #761, opened 2025-09-03
fix: update commons library to support java 8

venikkin:
Update auth token provider to support JVM 8

* update connectors-commons dependency to rc2
* downgrade keycloak container dependency conditionally for jdk 8
* replace `isBlank` with `isEmpty`

### DIFF
--- a/spark-3/src/test/scala/org/neo4j/spark/ReauthenticationIT.scala
+++ b/spark-3/src/test/scala/org/neo4j/spark/ReauthenticationIT.scala
@@ -27,7 +27,6 @@ import org.neo4j.driver.Driver
 import org.neo4j.driver.GraphDatabase
 import org.neo4j.spark.ReauthenticationIT.KEYCLOAK
 import org.neo4j.spark.ReauthenticationIT.NEO4J
-import org.neo4j.spark.SparkConnectorScalaSuiteIT.server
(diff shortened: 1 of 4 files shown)


spark-3/src/test/scala/org/neo4j/spark/ReauthenticationIT.scala
```diff
@@ -38,7 +37,12 @@ object ReauthenticationIT {
 
   private val KEYCLOAK = new KeycloakContainer("quay.io/keycloak/keycloak:26.2.5")
     .withNetwork(NETWORK).withNetworkAliases("keycloak")
-    .useTlsKeystore("/neo4j-keycloak.jks", "testpwd")
+    .withCopyFileToContainer(
```
fbiville, line 53: question: I assume this is because `useTlsKeystore` is not available in `1.9.0`, right?

fbiville (approved):

spark-3/src/test/scala/org/neo4j/spark/ReauthenticationIT.scala
venikkin: Yes, exactly. It was introduced after 1.9.0. However, it's easy to do with regular container methods.

PR #762, opened 2025-09-03
ci: make release announcement to team-spark too

ali-ince:
Upgrade jreleaser and use multiple channels for release announcements.


dhrudevalia (approved):
Is there a reason we aren't using the plugin?

venikkin (approved):

fbiville (approved):

ali-ince:
> Is there a reason we aren't using the plugin?

Yes, the yaml approach works much better since we build the connector twice, with scala 2.12 & 2.13, and update the maven coordinates accordingly. I think we can still make it work with the plugin, but it will be more compex to understand imho.

PR #773, opened 2025-09-12
feat: introduce skip.null.keys options

ali-ince:
This PR introduces a new option named `node.keys.skip.null.keys`, `relationship.keys.skip.nulls`, `relationship.source.node.keys.skip.nulls` and `relationship.target.node.keys.skip.nulls` which filters out creation of corresponding entities, whose defined `keys` properties contain a null value. 

These options will silently skip those incoming rows and avoids errors to be thrown due to this data problem.

### DIFF
--- a/common/src/main/scala/org/neo4j/spark/service/MappingService.scala
+++ b/common/src/main/scala/org/neo4j/spark/service/MappingService.scala
@@ -34,8 +34,6 @@ import org.neo4j.spark.util.Neo4jOptions
 import org.neo4j.spark.util.Neo4jUtil
 import org.neo4j.spark.util.QueryType
 import org.neo4j.spark.util.RelationshipSaveStrategy
-import org.neo4j.spark.util.ValidateSchemaOptions
-import org.neo4j.spark.util.Validations
 
 import java.util
 import java.util.function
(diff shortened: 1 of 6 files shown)


nord-erik (approved):

common/src/main/scala/org/neo4j/spark/service/MappingService.scala
```diff
@@ -66,7 +67,17 @@ class Neo4jWriteMappingStrategy(private val options: Neo4jOptions)
         }
       })
 
-    rowMap
+    if (options.skipNullKeys && keys.containsValue(Values.NULL)) {
```
sandhyaakula-pr: Where do we check if save mode is 'overwrite' ?

venikkin (commented):
The naming problem I can see is that we skip not all keys defined on the Neo4j side, but only those that an user marked as key for match query. Technically, user can have `skip.null.keys` and either use `append` mode ot even with `overwrite` mode and still receive a constraint error (correct me if I'm not wrong). 
We can explain this in the docs, but do you think we need to do any of those: 
* throw an error if `skip.null.keys` is used with other than `overwrite` modes
* (probably not, but want to hear your opinion), check the constraints and extend this check for all keys
?

common/src/main/scala/org/neo4j/spark/service/MappingService.scala
ali-ince: Good catch, added a validation for this case so this configuration will not be allowed at all.

ali-ince: added a validation in https://github.com/neo4j/neo4j-spark-connector/pull/773/commits/ab2dc05fa3ec85199b8b0edd1ed12e9a07942702.

ali-ince: I decided to remove this `Overwrite` restriction totally, so this no longer applies.

fbiville (approved):

PR #775, opened 2025-09-13
fix: apply transaction config to auto-commit transactions

venikkin:
When the default database timeout is smaller than the time a query can take, we need to apply transaction config to schema queries as well.

### DIFF
--- a/common/src/main/scala/org/neo4j/spark/service/SchemaService.scala
+++ b/common/src/main/scala/org/neo4j/spark/service/SchemaService.scala
@@ -435,14 +435,15 @@ class SchemaService(
     }
     log.info(s"Executing the following counting query on Neo4j: $query")
     session.readTransaction(
-      tx => tx.run(query, Values.value(Neo4jUtil.paramsFromFilters(filters).asJava)),
+      tx =>
+        tx.run(query, Values.value(Neo4jUtil.paramsFromFilters(filters).asJava))
+          .list()
+          .asScala
+          .map(_.get("count"))
+          .map(count => if (count.isNull) 0L else count.asLong())
+          .min,
       sessionTransactionConfig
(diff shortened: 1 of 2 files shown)


common/src/main/scala/org/neo4j/spark/service/SchemaService.scala
```diff
@@ -436,7 +434,10 @@ class SchemaService(
       queryReadStrategy.createStatementForNodeCount(options)
     }
     log.info(s"Executing the following counting query on Neo4j: $query")
-    session.run(query, Values.value(Neo4jUtil.paramsFromFilters(filters).asJava))
+    session.readTransaction(
```
ali-ince, line 437: is there a specific reason to switch to `readTransaction` from `run`?

venikkin: `run(query, value)` doesn't accept transaction config. I guess I can try replacing `Value` with a map.

ali-ince (approved):
Looks good. I think we should consider using `readTransaction` everywhere to get some driver level retry mechanism in place. Any thoughts @fbiville?

fbiville (approved):

common/src/main/scala/org/neo4j/spark/service/SchemaService.scala
fbiville: might be worth raising this missing overload to drivers

fbiville:
> Looks good. I think we should consider using `readTransaction` everywhere to get some driver level retry mechanism in place. Any thoughts @fbiville?

yep, `writeTransaction` is trickier but I agree that everywhere we used `session.run` to read, `readTransaction` would be a great replacement

PR #781, opened 2025-09-18
ci: introduce snyk scan step

venikkin:
**Problem**
Some maven profiles change dependencies, and we need to reflect this in our vulnerability scans.

**Solution**
Options viewed:
* Snyk UI - doesn't support custom maven parameters. 
* Snyk maven plugin - despite being a maven plugin, it doesn't respect maven profile properly and is basically a wrapper around Snyk CLI. 
* Snyk CLI - is capable of applying different maven arguments, so was chosen as a right approach 

Snyk CLI is integrated via a custom script TeamCity step. It's parallel to the main build-test flow and fails the result step of failure. 

**Drowbacks**
* This is not synchronised with the Snyk UI. Exclusions are managed via `.snyk` file. So we might need to exclude the same complaint twice in some cases. Because of this reason, I included only non-default maven profiles. 
* Doesn't do periodic scans. This is partially solved by periodical compatibility builds.

### DIFF
--- a/.teamcity/builds/Common.kt
+++ b/.teamcity/builds/Common.kt
@@ -127,6 +127,12 @@ enum class Neo4jVersion(val version: String, val dockerImage: String) {
   ),
 }
 
+data class SnykProfile(val name: String, val mavenArgs: String, val dockerImage: String = "snyk/snyk:maven-3-jdk-17")
+
+val SNYK_PROFILES = setOf(
+    SnykProfile("scala-2-13", "-Pscala-2.13")
+)
+
 fun <S, T, Y> Iterable<S>.cartesianProduct(
     other1: Collection<T>,
     other2: Collection<Y>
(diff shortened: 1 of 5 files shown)


.teamcity/builds/Common.kt
```diff
@@ -127,6 +127,13 @@ enum class Neo4jVersion(val version: String, val dockerImage: String) {
   ),
 }
 
+data class SnykProfile(val name: String, val mavenArgs: String, val dockerImage: String = "snyk/snyk:maven-3-jdk-17")
+
+val SNYK_PROFILES = setOf(
+    SnykProfile("scala-2-13", "-Pscala-2.13"),
+    SnykProfile("neo4j-4-4", "-Pneo4j-4.4"),
```
ali-ince: Do we need `neo4j-4-4` profile at all, since it doesn't affect the packaged artifacts for the connector?

ali-ince (approved):

.teamcity/builds/Common.kt
venikkin: You are right. We don't need this. Removing...

PR #790, opened 2025-10-08
feat: introduce global await for index creation

nord-erik:
Right before data is being written and right after scripts have been invoked we now  call
cypher procedure "db.awaitIndexes".

This ensures that any custom indexes created from custom scripts will have leeway room to
finish construction before we write data.

Can be tweaked using option "index.await.timeout". Default 300, set to 0 for old behaviour
i.e. no wait.

### DIFF
--- a/common/src/main/scala/org/neo4j/spark/util/Neo4jOptions.scala
+++ b/common/src/main/scala/org/neo4j/spark/util/Neo4jOptions.scala
@@ -134,6 +134,8 @@ class Neo4jOptions(private val options: java.util.Map[String, String]) extends S
     )
   }
 
+  val indexAwait = getParameter(INDEX_AWAIT_TIMEOUT_SEC, DEFAULT_INDEX_AWAIT_TIMEOUT_SEC.toString).toInt
+
   val query: Neo4jQueryOptions = (
     getParameter(QUERY.toString.toLowerCase),
     getParameter(LABELS.toString.toLowerCase),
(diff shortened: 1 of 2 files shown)


neo4j-connectors:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>:exclamation: Please include a description of your PR changes.</td>
    </tr>
  </tbody>
</table>


> Pull request should have a description of the underlying changes.
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against c467ad48cdb4beb50687b09f9cf6edd0660ac497
</p>

common/src/main/scala/org/neo4j/spark/util/Neo4jOptions.scala
```diff
@@ -647,6 +652,8 @@ object Neo4jOptions {
 
   val DEFAULT_MAP_GROUP_DUPLICATE_KEYS = false
 
+  val DEFAULT_INDEX_AWAIT_TIMEOUT_MSEC = 300
```
sandhyaakula-pr: nit : MSEC meant milliseconds here? I suppose its 'seconds' according to the [documentation](https://neo4j.com/docs/operations-manual/current/procedures/#procedure_db_awaitIndexes).

nord-erik: thank you so much, I'll update

venikkin (approved):
Looks good to me! Regarding testing, timeout is indeed hard to test. I think the best we can do is check query log or driver log that the valid statement is requested and rely on the database for the rest.

nord-erik:
I've managed to do a manual test setup, but need to further confirm. When calling db.awaitIndexes() after creating constraints, there seem to be a deadlock of some kind because awaitIndexes call completely blocks constraint creation. Will not merge yet

update: this does not happen in the spark connector itself

nord-erik:
draft docs: https://github.com/neo4j/docs-spark/pull/83

PR #791, opened 2025-10-08
feat: import exactly relationship.properties when specified

venikkin:
Changes the behaviour of `relationship.properties` the following way:
* if set explicitly, imports listed fields as properties (of set to empty string, imports none fields)
* if omitted, preserves old behaviour, which is importing all possible fields except used for node keys


nord-erik (approved):
lgtm

PR #816, opened 2025-12-10
feat: introduce byte, short, decimal, timestampNTZ types

nord-erik:
adds support to natively write these spark types to neo4j:

- Byte: will convert to Integer, therefore cannot read-back as Byte
- Short: will convert to Integer, therefore cannot read-back as Short
- Decimal: will convert to String, therefore cannot read-back as Decimal
- Binary: will convert to ByteArray and will be read-back as ByteArray
- TimestampNTZType: will convert to LocalDateTime

BREAKING CHANGE:

- already existing Timestamp now converts to ZonedDateTime, not LocalDateTime.

Before, writing a Timestamp would remove the timezone information.
Now it will be normalized to UTC-time instead. If you want to retain old behaviour,
please use TimestampNTZType or be mindful when you read the stamp.

Assuming you interpret time zone when reading and writing, you should not notice.
But if you indeed intend to use LocalDateTime, this is moved and maps to TimestampNTZType


neo4j-connectors:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>:exclamation: Please include a description of your PR changes.</td>
    </tr>
  </tbody>
</table>


> Pull request should have a description of the underlying changes.
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against e851e20759cb313af592a1eaa27bb050510a9d56
</p>

.snyk
```diff
@@ -16,4 +16,5 @@ ignore:
   'SNYK-JAVA-ORGGLASSFISHJERSEYCORE-14049172': *spark-core-exclusions
   'SNYK-JAVA-ORGLZ4-14151788': *spark-core-exclusions
   'SNYK-JAVA-ORGLZ4-14219384': *spark-core-exclusions
+  'SNYK-JAVA-IOAIRLIFT-14412703': *spark-core-exclusions
```
nord-erik: I will remove this, just temporarily unblock, if we want this in, should be separated isolated PR

nord-erik: Ok did not bite

neo4j-connectors:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>Commit Message 'fix: LocalDate and Date point towards same tye': subject must not be sentence-case, start-case, pascal-case, upper-case</td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 3319cce1b41b41d85d1d4590aab44d7d782dabda
</p>

common/src/main/scala/org/neo4j/spark/converter/DataConverter.scala
```diff
@@ -77,29 +77,23 @@ object SparkToNeo4jDataConverter {
...
+        SparkToNeo4jDataConverter.yearMonthIntervalToNeo4j(intValue)
       case longValue: Long if dataType == DataTypes.TimestampType =>
-        convert(
-          DateTimeUtils
-            .toJavaTimestamp(longValue),
-          dataType
-        )
+        convert(DateTimeUtils.toJavaTimestamp(longValue), dataType)
+      case longValue: Long if dataType == DataTypes.TimestampNTZType =>
+        convert(DateTimeUtils.microsToLocalDateTime(longValue), dataType)
```
fbiville, line 92: question: why micros? the docs only mention second precision

common/src/main/scala/org/neo4j/spark/converter/DataConverter.scala
```diff
@@ -77,29 +77,23 @@ object SparkToNeo4jDataConverter {
...
+      case longValue: Long if dataType.isInstanceOf[DayTimeIntervalType] =>
+        SparkToNeo4jDataConverter.dayTimeIntervalToNeo4j(longValue)
```
fbiville: nit: I know it's an existing method, but the name is not quite clear in terms of output type

common/src/main/scala/org/neo4j/spark/converter/DataConverter.scala
```diff
@@ -77,29 +77,23 @@ object SparkToNeo4jDataConverter {
...
-        return SparkToNeo4jDataConverter.dayTimeIntervalToNeo4j(value.asInstanceOf[Long])
-      case _: YearMonthIntervalType if value != null =>
-        return SparkToNeo4jDataConverter.yearMonthIntervalToNeo4j(value.asInstanceOf[Int])
-      case _ => // do nothing
-    }
-
     value match {
       case date: java.sql.Date           => convert(date.toLocalDate, dataType)
-      case timestamp: java.sql.Timestamp => convert(timestamp.toLocalDateTime, dataType)
+      case timestamp: java.sql.Timestamp => convert(timestamp.toInstant.atZone(ZoneOffset.UTC), dataType)
```
fbiville, line 80: is there a way to have something like:
```
case timestamp: java.sql.Timestamp if dataType == DataTypes.TimestampType => convert(timestamp.toInstant.atZone(ZoneOffset.UTC), dataType)
case timestamp: java.sql.Timestamp if dataType == DataTypes.TimestampNTZType => convert(timestamp.toLocalDateTime, dataType)
```
?
That would potentially limit the blast radius of the breaking change (timestamps without time zone should still be translated to local date times)

common/src/main/scala/org/neo4j/spark/converter/TypeConverter.scala
```diff
@@ -125,26 +128,35 @@ object SparkToCypherTypeConverter {
...
+    DataTypes.createArrayType(DataTypes.ShortType, false) -> "LIST<INTEGER NOT NULL>",
     DataTypes.createArrayType(DataTypes.IntegerType, false) -> "LIST<INTEGER NOT NULL>",
     DataTypes.createArrayType(DataTypes.LongType, false) -> "LIST<INTEGER NOT NULL>",
     DataTypes.createArrayType(DataTypes.FloatType, false) -> "LIST<FLOAT NOT NULL>",
     DataTypes.createArrayType(DataTypes.DoubleType, false) -> "LIST<FLOAT NOT NULL>",
     DataTypes.createArrayType(DataTypes.DateType, false) -> "LIST<DATE NOT NULL>",
-    DataTypes.createArrayType(DataTypes.TimestampType, false) -> "LIST<LOCAL DATETIME NOT NULL>",
-    DataTypes.createArrayType(DataTypes.TimestampType, true) -> "LIST<LOCAL DATETIME NOT NULL>",
+    DataTypes.createArrayType(DataTypes.TimestampType, false) -> "LIST<ZONED DATETIME NOT NULL>",
+    DataTypes.createArrayType(DataTypes.TimestampType, true) -> "LIST<ZONED DATETIME NOT NULL>",
```
fbiville: question: we cannot really have null zoned datetimes in lists, can we?

common/src/main/scala/org/neo4j/spark/converter/TypeConverter.scala
```diff
@@ -125,26 +128,35 @@ object SparkToCypherTypeConverter {
...
+    DataTypes.createArrayType(DataTypes.TimestampNTZType, false) -> "LIST<LOCAL DATETIME NOT NULL>",
+    DataTypes.createArrayType(DataTypes.TimestampNTZType, true) -> "LIST<LOCAL DATETIME NOT NULL>",
```
fbiville: question: we cannot really have null local datetimes in lists, can we?

common/src/main/scala/org/neo4j/spark/converter/DataConverter.scala
fbiville: it's such a pity they used `java.sql.Timestamp` for timestamps with time zones -_-'

test-support/src/main/scala/org/neo4j/spark/SparkConnectorScalaBaseTSE.scala
```diff
@@ -23,6 +23,9 @@ import org.junit.rules.TestName
 import org.neo4j.Closeables.use
 import org.scalatestplus.junit.AssertionsForJUnit
 
+import java.time.ZoneId
+import java.util.TimeZone
```
fbiville: nit: unused imports?

common/src/main/scala/org/neo4j/spark/converter/DataConverter.scala
nord-erik: I reckon the spark-side is microseconds:

https://api-docs.databricks.com/scala/spark/latest/org/apache/spark/sql/types/TimestampNTZType.html

common/src/main/scala/org/neo4j/spark/converter/DataConverter.scala
nord-erik: I'll have a look

common/src/main/scala/org/neo4j/spark/converter/TypeConverter.scala
nord-erik: I'd assume so as well. No idea what that would look like / represent like if it was the case

common/src/main/scala/org/neo4j/spark/converter/DataConverter.scala
fbiville: ah, https://spark.apache.org/docs/latest/sql-ref-datatypes.html is not up-to-date, then.

common/src/main/scala/org/neo4j/spark/converter/TypeConverter.scala
fbiville: in that case, we cannot call `createArray` with `true`. Afaik, Cypher does not support `null`s in lists.

common/src/main/scala/org/neo4j/spark/converter/DataConverter.scala
nord-erik: At least in tests we never touch the case of java.sql.Timestamp && dataType = DataTypes.TimestampNTZType

Spark itself will only associate these ones:
TimestampType <-> java.sql.Timestamp
TimestampNTZType <-> java.time.LocalDateTime

So we will never really get these associations from Spark natively:
TimestampType <-> java.time.LocalDateTime
TimestampNTZType <-> java.sql.Timestamp

Before this PR, what we did, was that WE converted:
TimestypeType -> java.sql.Timestamp -> java.time.LocalDateTime

this is now removed, "fixing the bug". Still learning as I go but quite confident the above is true, good to try poke holes in this before we merge

common/src/main/scala/org/neo4j/spark/converter/TypeConverter.scala
nord-erik: hmm... confused about this to be honest. when I did durations I added both {true, false} because timestamps already had it.

what's the consequences here to remove? I'll remove and see if all tests pass in the meantime.

fbiville (approved):

PR #796, opened 2025-10-28
feat: native support for intervals <-> Neo4j duration

nord-erik:
this commit introduces a new way to write from Spark SQL interval types to Neo4j duration type.
commit is additive, i.e. the previous method to write Neo4j duration type via custom struct is
still possible, and should therefore be backwards compatible.

part of CONN-341

PR #836, opened 2026-01-08
feat: allow users to enable the legacy timestamp conversion

fbiville:
This brings back the legacy type conversion logic of supported types that was replaced in #816 and #796

### DIFF
--- a/common/src/main/scala/org/neo4j/spark/converter/DataConverter.scala
+++ b/common/src/main/scala/org/neo4j/spark/converter/DataConverter.scala
@@ -76,10 +76,19 @@ object SparkToNeo4jDataConverter {
 class SparkToNeo4jDataConverter(options: Neo4jOptions) extends DataConverter[Value] {
 
   override def convert(value: Any, dataType: DataType): Value = {
+    if (options.legacyTypeConversionEnabled) {
+      dataType match {
+        case _: DayTimeIntervalType if value != null =>
+          return SparkToNeo4jDataConverter.dayTimeMicrosToNeo4jDuration(value.asInstanceOf[Long])
+        case _: YearMonthIntervalType if value != null =>
+          return SparkToNeo4jDataConverter.yearMonthIntervalToNeo4jDuration(value.asInstanceOf[Int])
+        case _ => // do nothing
+      }
+    }
     value match {
       case date: java.sql.Date => convert(date.toLocalDate, dataType)
       case timestamp: java.sql.Timestamp =>
-        if (options.legacyTimestampConversionEnabled) {
+        if (options.legacyTypeConversionEnabled) {
           convert(timestamp.toLocalDateTime, dataType)
         } else {
           convert(timestamp.toInstant.atZone(ZoneOffset.UTC), dataType)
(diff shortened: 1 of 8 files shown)


common/src/main/scala/org/neo4j/spark/util/Neo4jOptions.scala
```diff
@@ -621,6 +626,9 @@ object Neo4jOptions {
 
   val SCRIPT = "script"
 
+  // Data conversion
+  val LEGACY_TIMESTAMP_CONVERSION_ENABLED = "conversion.timestamp.legacy_enabled"
```
ali-ince: by looking to other options, should we name this as `conversion.timestamp.legacy.enabled`?

common/src/main/scala/org/neo4j/spark/converter/TypeConverter.scala
```diff
@@ -153,17 +162,35 @@ object SparkToCypherTypeConverter {
...
+
+  private def mapping(sourceType: DataType, options: Neo4jOptions): String = {
+    val mappings = sourceTypeMappings(options)
+    mappings(sourceType)
+  }
+
+  private def sourceTypeMappings(options: Neo4jOptions): Map[DataType, String] = {
+    var result = baseMappings
+    if (options.legacyTimestampConversionEnabled) {
+      result += (DataTypes.TimestampType -> "LOCAL DATETIME")
```
ali-ince, line 174: I see previously we had two entries `TimestampType` & `TimestampNTZType` where here we have only one. Is that intentional?

common/src/main/scala/org/neo4j/spark/converter/DataConverter.scala
```diff
@@ -88,7 +94,7 @@ class SparkToNeo4jDataConverter extends DataConverter[Value] {
         SparkToNeo4jDataConverter.yearMonthIntervalToNeo4jDuration(intValue)
       case longValue: Long if dataType == DataTypes.TimestampType =>
         convert(DateTimeUtils.toJavaTimestamp(longValue), dataType)
-      case longValue: Long if dataType == DataTypes.TimestampNTZType =>
+      case longValue: Long if dataType == DataTypes.TimestampNTZType && !options.legacyTimestampConversionEnabled =>
```
ali-ince: and what happens if the we are not on legacy code path?

common/src/main/scala/org/neo4j/spark/converter/DataConverter.scala
```diff
@@ -217,8 +223,14 @@ class Neo4jToSparkDataConverter extends DataConverter[Any] {
           ))
         }
         case zt: ZonedDateTime => DateTimeUtils.instantToMicros(zt.toInstant)
-        case dt: LocalDateTime => DateTimeUtils.localDateTimeToMicros(dt)
-        case d: LocalDate      => d.toEpochDay.toInt
+        case dt: LocalDateTime => {
```
ali-ince, line 226: maybe I got confused but, shouldn't the conversion logic be inversed in this if block?

common/src/main/scala/org/neo4j/spark/converter/TypeConverter.scala
fbiville: there was no support for `TimestampNTZType` prior to #816

common/src/main/scala/org/neo4j/spark/converter/DataConverter.scala
fbiville: if legacy is **not** enabled, then this particular path is active.
if it is enabled, then there's no explicit support for `TimestampNTZType` just like it was prior to #816

common/src/main/scala/org/neo4j/spark/converter/DataConverter.scala
fbiville: `options.legacyTimestampConversionEnabled` means "logic is equivalent to what was before #816"
see https://github.com/neo4j/neo4j-spark-connector/pull/816/changes#diff-e571424377b7f54a2916ea748642926b6ef966b9556e91e98ebb2cd289cf8c0cL223

common/src/main/scala/org/neo4j/spark/util/Neo4jOptions.scala
fbiville: I thought about it, but I'm not sure `legacy` stands on its own... would we envision `conversion.timestamp.legacy.<something else>`?

nord-erik:
functionally I think this looks great now :100:

common/src/main/scala/org/neo4j/spark/util/Neo4jOptions.scala
fbiville: we settled to `type.conversion=(default|legacy)` in Slack

ali-ince (approved):
LGTM

PR #838, opened 2026-01-09
test: update wait strategy for retries

ali-ince:
This PR simplifies and fixes the wait strategy to be retried during the startup time period. Hopefully will resolve the transaction timeout errors in tests completely.

### DIFF
--- a/test-support/src/main/scala/org/neo4j/Neo4jContainerExtension.scala
+++ b/test-support/src/main/scala/org/neo4j/Neo4jContainerExtension.scala
@@ -53,7 +53,7 @@ class DatabasesWaitStrategy(private val auth: AuthToken) extends AbstractWaitStr
             val session = driver.session(SessionConfig.forDatabase("system"))
             try {
               databases.foreach { db =>
-                session.writeTransaction(tx => tx.run(s"CREATE DATABASE $db IF NOT EXISTS WAIT").consume())
+                session.writeTransaction(tx => tx.run(s"CREATE DATABASE $db IF NOT EXISTS WAIT 30 SECONDS").consume())
 
                 val status = session.readTransaction(tx => {
                   tx.run(s"SHOW DATABASE $db YIELD currentStatus").single().get("currentStatus").asString()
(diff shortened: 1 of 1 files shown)


test-support/src/main/scala/org/neo4j/Neo4jContainerExtension.scala
```diff
@@ -44,51 +44,31 @@ class DatabasesWaitStrategy(private val auth: AuthToken) extends AbstractWaitStr
...
-                    throw new RuntimeException(s"Cannot started because of the following databases: ${notOnline.keys}")
-                  }
+        new Callable[Unit] {
+          override def call(): Unit = {
+            val session = driver.session(SessionConfig.forDatabase("system"))
+            try {
+              databases.foreach { db =>
+                session.writeTransaction(tx => tx.run(s"CREATE DATABASE $db IF NOT EXISTS WAIT").consume())
+
+                val status = session.readTransaction(tx => {
```
fbiville, line 58: question: is this still needed? Doesn't the `WAIT` clause make the DB creation query execution completely synchronous?

test-support/src/main/scala/org/neo4j/Neo4jContainerExtension.scala
```diff
@@ -44,51 +44,31 @@ class DatabasesWaitStrategy(private val auth: AuthToken) extends AbstractWaitStr
...
-
-                  if (databasesStatus.size < databases.size || notOnline.nonEmpty) {
-                    throw new RuntimeException(s"Cannot started because of the following databases: ${notOnline.keys}")
-                  }
+        new Callable[Unit] {
+          override def call(): Unit = {
+            val session = driver.session(SessionConfig.forDatabase("system"))
+            try {
+              databases.foreach { db =>
+                session.writeTransaction(tx => tx.run(s"CREATE DATABASE $db IF NOT EXISTS WAIT").consume())
```
fbiville: question: shall we add a timeout?

test-support/src/main/scala/org/neo4j/Neo4jContainerExtension.scala
ali-ince: nope, `WAIT` is best effort as far as I remember.

ali-ince: see https://neo4j.com/docs/operations-manual/current/database-administration/standard-databases/wait-options/.

test-support/src/main/scala/org/neo4j/Neo4jContainerExtension.scala
ali-ince: adding a `30 SECONDS`.

fbiville (approved):

PR #876, opened 2026-04-14
fix: type constraints prioritise target property name

nord-erik:
There is a bug where creating type constraints would pick source names.
This makes it so that if a client uses a custom mapping, the constraint
will not apply.	        

This commit fixes that for options:
- node.keys
- relationship.properties

Added test cases for remapped situations, asserting existance and type   
constraints.

### DIFF
--- a/spark-3/src/test/scala/org/neo4j/spark/DataSourceSchemaWriterTSE.scala
+++ b/spark-3/src/test/scala/org/neo4j/spark/DataSourceSchemaWriterTSE.scala
@@ -62,25 +62,27 @@ class DataSourceSchemaWriterTSE extends SparkConnectorScalaBaseTSE {
 
   final private val SHOW_CONSTRAINTS_QUERY =
     """|SHOW CONSTRAINTS
-       |YIELD name, type, entityType, labelsOrTypes, properties, ownedIndex""".stripMargin
+       |YIELD name, type, entityType, labelsOrTypes, properties, ownedIndex, propertyType""".stripMargin
 
   final private val NODE_UNIQUENESS_SHOW_CONSTRAINTS_QUERY =
     """|SHOW CONSTRAINTS
-       |YIELD name, type AS ptype, entityType, labelsOrTypes, properties, ownedIndex
+       |YIELD name, type AS ptype, entityType, labelsOrTypes, properties, ownedIndex, propertyType
        |RETURN name, entityType, labelsOrTypes, properties, ownedIndex,
        |CASE ptype
        |  WHEN "UNIQUENESS" THEN "NODE_PROPERTY_UNIQUENESS"
        |  ELSE ptype
-       |END AS type""".stripMargin
+       |END AS type, propertyType
+       |ORDER BY type ASC""".stripMargin
 
   final private val RELATIONSHIP_UNIQUENESS_SHOW_CONSTRAINTS_QUERY =
     """|SHOW CONSTRAINTS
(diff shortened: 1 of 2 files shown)


spark-3/src/test/scala/org/neo4j/spark/DataSourceSchemaWriterTSE.scala
```diff
@@ -128,399 +150,211 @@ class DataSourceSchemaWriterTSE extends SparkConnectorScalaBaseTSE {
...
+      .map(_.asMap(v => v.asObject()).asScala.mapValues(mapData).toMap)
+      .toSeq
+
+    assertEquals(expectedSchema, actualSchema)
 
-    assertEquals(expectedNode, actualNode)
+    val expectedNode =
+      node.map {
+        case (k, v) =>
+          if (k == "string" || k == "int") (k + "_prop", v)
```
fbiville, line 331: nit: you can declare a map for the mappings, and create the `node.keys` setting from it and in `map` right here

spark-3/src/test/scala/org/neo4j/spark/DataSourceSchemaWriterTSE.scala
```diff
@@ -128,399 +150,211 @@ class DataSourceSchemaWriterTSE extends SparkConnectorScalaBaseTSE {
...
+          else (k, v)
+      }
+
+    assertEquals(expectedNode, actualNodeWithSchema())
```
fbiville: nit: I'd actually inline `actualNodeWithSchema()` and leave the duplicates, and extracting a function here hurts the test assertion readability imo (please ignore if you don't agree)

spark-3/src/test/scala/org/neo4j/spark/DataSourceSchemaWriterTSE.scala
```diff
@@ -936,13 +716,76 @@ class DataSourceSchemaWriterTSE extends SparkConnectorScalaBaseTSE {
...
+      .option("url", SparkConnectorScalaSuiteIT.server.getBoltUrl)
+      .option("labels", ":SurnameUnique")
+      .option("node.keys", "surname:surname_unique")
+      .option(Neo4jOptions.SCHEMA_OPTIMIZATION_NODE_KEY, ConstraintsOptimizationType.UNIQUE.toString)
+      .save()
+
+    val actualConstraint = SparkConnectorScalaSuiteIT.session().run(NODE_UNIQUENESS_SHOW_CONSTRAINTS_QUERY)
+      .list()
+      .asScala
+      .map(_.asMap(v => v.asObject()).asScala.mapValues(mapData).filter(k => k._1 != "id").toMap)
```
fbiville: question: why do you need to filter out `id` here? there's no `id` property in the test, is there?

spark-3/src/test/scala/org/neo4j/spark/DataSourceSchemaWriterTSE.scala
nord-erik: yeah I went back and forth on this one, can revert to verbose but duplicated

spark-3/src/test/scala/org/neo4j/spark/DataSourceSchemaWriterTSE.scala
nord-erik: yeah I think you're right. it's copy-paste slop from line 711 (existing code also redundantly filter on id)

fbiville (approved):

PR #877, opened 2026-04-15
chore: remove alias from dependabot

nord-erik:
error from github UI:

```
Dependabot couldn't parse the config file. The error raised was:

YAML aliases are not supported
```

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,12 +6,19 @@ updates:
       interval: "daily"
     cooldown:
       default-days: 3
-  - &maven
-    package-ecosystem: "maven"
+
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -6,12 +6,19 @@ updates:
       interval: "daily"
     cooldown:
       default-days: 3
-  - &maven
-    package-ecosystem: "maven"
+
+  - package-ecosystem: "maven"
     directory: "/"
+    target-branch: "5.0"
```
nord-erik, line 12: explicit mention of branch, so when we swap github's default, we still point to 5.0

fbiville (approved):

PR #897, opened 2026-04-22
refactor: simplify project structure

Emrehzl94:
This PR: 
- Merges all modules into a single root module. All sources now live directly under src/ in the project root.                                                              
- Updates build maven-release.sh and CI pipeline accordingly.
- Renames maven coordinates to org.neo4j.connectors:spark (was org.neo4j:neo4j-connector-apache-spark)
- Overwrites maven version instead of artifactId for each Scala version.
- Renames final build neo4j-spark-connector-{projectVersion}-s_{scalaVersion}\_for_spark_{sparkVersion}.jar

### Important
CI is failing because TeamCity is still using the config from the main branch, where it expects artifacts in spark/target/. But after the refactor, the JARs are now in target/.
The path is fixed in .teamcity/builds/Common.kt:
`spark/target/*_for_spark_*.jar → target/*_for_spark_*.jar`
Once this PR is merged and TeamCity picks up the updated config, it should start finding the artifacts again and the CI should pass.


pom.xml
```diff
@@ -554,6 +680,47 @@
...
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-surefire-plugin</artifactId>
+            </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-resources-plugin</artifactId>
+                <version>${maven-resources-plugin.version}</version>
+                <configuration>
+                    <encoding>ISO-8859-1</encoding>
```
fbiville: question: probably not new, but why not UTF-8? 🤔

Emrehzl94: tbh have no idea, just copied from the original

nord-erik (approved):

PR #907, opened 2026-05-06
chore: update husky scripts

ali-ince:
This pull request introduces improvements to the pre-commit hook.


ali-ince:
Merging this as this is very similar to #908.

PR #930, opened 2026-05-19
feat: expose cypher query tuning parameters

nord-erik:
This PR exposes cypher query tuning parameters and allows users to
specifiy these parameters.

Supports all non-deprecated options that is supported by Neo4j 5.0
or later. Full list of supported parameters:
- expressionEngine
- inferSchemaParts
- interpretedPipesFallback
- operatorEngine
- planner
- replan
- runtime
- updateStrategy

You specify these by passing them as Spark Options prefixed with
cypher like so

```
.option("labels", ":Person")
.option("cypher.runtime", "parallel")
.option("cypher.operator.engine", "compiled")
```


src/main/scala/org/neo4j/spark/util/Neo4jOptions.scala
```diff
@@ -662,6 +684,16 @@ object Neo4jOptions {
...
 
+  // custom cypher tuning parameters
+  val TUNING_EXPRESSION_ENGINE = "tuning.expression.engine"
+  val TUNING_INFER_SCHEMA_PARTS = "tuning.infer.schema.parts"
+  val TUNING_INTERPRETED_PIPES_FALLBACK = "tuning.interpreted.pipes.fallback"
+  val TUNING_OPERATOR_ENGINE = "tuning.operator.engine"
+  val TUNING_PLANNER = "tuning.planner"
+  val TUNING_REPLAN = "tuning.replan"
+  val TUNING_RUNTIME = "tuning.runtime"
+  val TUNING_UPDATE_STRATEGY = "tuning.update.strategy"
```
nord-erik: happy to hear suggested names, not super happy with tuning

but it references:
https://neo4j.com/docs/cypher-manual/current/planning-and-tuning/query-tuning

without conflicting with existing "query" options.

fbiville: Maybe `cypher` instead of `tuning`?

nord-erik:
Legends:
V = version preamble is present
T = tuning preamble is present
! = not supported
_ = not present

So e.g. "V _" version preamble is there, but not tuning


## Before
```
		nodes	rel		query	gds
read	V _		V _		_ _		V _
write	V _		V _		_ _		! !

```

## After
```

		nodes	rel		query	gds
read	V T		V T		V T		V !
write	V T		V T		V T		! !
```

src/main/scala/org/neo4j/spark/util/Neo4jOptions.scala
nord-erik: Yeah that works!

src/main/scala/org/neo4j/spark/service/Neo4jQueryService.scala
```diff
@@ -145,18 +154,22 @@ class Neo4jQueryWriteStrategy(private val neo4j: Neo4j, private val saveMode: Sa
...
 
 class Neo4jQueryReadStrategy(
-  neo4j: Neo4j,
-  filters: Array[Filter] = Array.empty[Filter],
-  partitionPagination: PartitionPagination = PartitionPagination.EMPTY,
-  requiredColumns: Seq[String] = Seq.empty,
-  aggregateColumns: Array[AggregateFunc] = Array.empty,
-  jobId: String = ""
+  private val neo4j: Neo4j,
+  private val tuningOptions: Neo4jTuningOptions,
```
Emrehzl94, line 158: Not a scala expert but, do we need this param here since we have the options from Neo4jQueryService?

src/main/scala/org/neo4j/spark/service/Neo4jQueryService.scala
```diff
@@ -495,6 +509,10 @@ class Neo4jQueryReadStrategy(
   }
 
   override def createStatementForGDS(options: Neo4jOptions): String = {
+    if (options.tuning != Neo4jTuningOptions.empty) {
+      throw new UnsupportedOperationException("Query tuning parameters are not supported for GDS queries")
+    }
+
```
Emrehzl94, line 515: should we abort the mission here? or maybe we could print out a warning and ignore tuning options. wdyt?

src/main/scala/org/neo4j/spark/service/Neo4jQueryService.scala
```diff
@@ -524,6 +542,28 @@ class Neo4jQueryReadStrategy(
...
+      partitionPagination,
+      requiredColumns,
+      aggregateColumns,
+      jobId
+    ) {
+
+  override def createStatementForQuery(options: Neo4jOptions): String = {
+    super.createStatementForQuery(options, false)
+  }
+}
```
Emrehzl94, line 565: when do we need this?

src/main/scala/org/neo4j/spark/service/Neo4jQueryService.scala
nord-erik: good idea, I think printing a warning and skip can be an improvement! (also documenting behaviour that it will skip)

src/main/scala/org/neo4j/spark/service/Neo4jQueryService.scala
nord-erik: We need this for custom queries, when this is passed: `.option("query", "YOUR QUERY HERE")`

These queries can sometimes be embedded inside a cypher call.
For that cypher call to be valid, the preamble needs to be outside.

So we need a mechanism for creating the cypher-template without the preamble.

I experimented with:
1) Adding the preamble inside the call, then parse it, then move it out using string manipulation
2) Having each code-path that send queries over the driver prepend preamble just-in-time

skipped 1) because it's needlessly complex
skipped 2) because then we need to track all places we do this on if we change preable in the future

nord-erik: concrete exmaple, this is how a query can look before the change (schema query):

```
CALL {
  Inner query goes here
} RETURN *
ORDER BY rand()
LIMIT 100
```

And this is invalid:

```
CALL {
  CYPHER runtime=parallel
  CYPHER 5
  Inner query goes here
} RETURN *
ORDER BY rand()
LIMIT 100
```

while this is valid (and implemented):

```
CYPHER runtime=parallel
CYPHER 5
CALL {
  Inner query goes here
} RETURN *
ORDER BY rand()
LIMIT 100
```

Emrehzl94 (approved):

PR #933, opened 2026-05-22
feat: allow override of cypher version as option

nord-erik:
introduce a new way to specify what `CYPHER <version>` will be prefixed
to queries. before we simply picked version `5` if the target neo4j
instance supported the syntax.

new behaviour is to read option `cypher.version`, validate the option
via caniuse. If set, it, will override old behaviour and prefix
queries with `CYPHER 5` or `CYPHER 25` respectively.

updates `CypherRenderer` to also respect the override when
picking dialect.

retains old behaviour if value is unset or empty. lastly, will throw
if the caniuse validation is not successful (bad input from user).

### DIFF
--- a/src/main/scala/org/neo4j/spark/cypher/CypherPreamble.scala
+++ b/src/main/scala/org/neo4j/spark/cypher/CypherPreamble.scala
@@ -19,13 +19,14 @@ package org.neo4j.spark.cypher
 import org.neo4j.caniuse.CanIUse.INSTANCE.canIUse
 import org.neo4j.caniuse.Cypher.{INSTANCE => Cypher}
 import org.neo4j.caniuse.Neo4j
+import org.neo4j.spark.util.Neo4jOptions
 import org.neo4j.spark.util.Neo4jTuningOptions
 
 object CypherPreamble {
 
-  def fullPreamble(neo4j: Neo4j, tuningOptions: Neo4jTuningOptions): String = {
-    val version = versionPreamble(neo4j)
-    val tuning = tuningPreamble(tuningOptions)
+  def fullPreamble(neo4j: Neo4j, neo4jOptions: Neo4jOptions): String = {
+    val version = versionPreamble(neo4j, neo4jOptions)
+    val tuning = tuningPreamble(neo4jOptions.tuning)
 
     if (tuning.isEmpty) {
       s"$version"
@@ -34,7 +35,15 @@ object CypherPreamble {
     }
   }
 
-  private def versionPreamble(neo4j: Neo4j): String = {
+  private def versionPreamble(neo4j: Neo4j, neo4jOptions: Neo4jOptions): String = {
+    if (neo4jOptions.cypherVersion == "5") {
+      return "CYPHER 5 "
+    }
+
+    if (neo4jOptions.cypherVersion == "25") {
+      return "CYPHER 25 "
+    }
+
     if (canIUse(Cypher.explicitCypher5Selection()).withNeo4j(neo4j)) {
       "CYPHER 5 "
     } else {
--- a/src/main/scala/org/neo4j/spark/cypher/CypherPreamble.scala
+++ b/src/main/scala/org/neo4j/spark/cypher/CypherPreamble.scala
@@ -36,12 +36,8 @@ object CypherPreamble {
   }
 
   private def versionPreamble(neo4j: Neo4j, neo4jOptions: Neo4jOptions): String = {
-    if (neo4jOptions.cypherVersion == "5") {
-      return "CYPHER 5 "
-    }
-
-    if (neo4jOptions.cypherVersion == "25") {
-      return "CYPHER 25 "
+    if (neo4jOptions.cypherVersion != null && neo4jOptions.cypherVersion.nonEmpty) {
+      return s"CYPHER ${neo4jOptions.cypherVersion} "
     }
 
     if (canIUse(Cypher.explicitCypher5Selection()).withNeo4j(neo4j)) {
--- a/src/main/scala/org/neo4j/spark/service/Neo4jQueryService.scala
+++ b/src/main/scala/org/neo4j/spark/service/Neo4jQueryService.scala
@@ -27,7 +27,6 @@ import org.apache.spark.sql.sources.Filter
 import org.apache.spark.sql.sources.Or
 import org.neo4j.caniuse.Neo4j
 import org.neo4j.cypherdsl.core._
-import org.neo4j.cypherdsl.core.renderer.Renderer
 import org.neo4j.spark.cypher.CypherPreamble.fullPreamble
 import org.neo4j.spark.cypher.CypherRenderer
 import org.neo4j.spark.util.Neo4jImplicits._
@@ -153,6 +152,7 @@ class Neo4jQueryWriteStrategy(
 
 class Neo4jQueryReadStrategy(
   private val neo4j: Neo4j,
+  private val renderer: CypherRenderer,
   private val filters: Array[Filter] = Array.empty[Filter],
   private val partitionPagination: PartitionPagination = PartitionPagination.EMPTY,
   private val requiredColumns: Seq[String] = Seq.empty,
@@ -160,7 +160,6 @@ class Neo4jQueryReadStrategy(
   private val jobId: String = "",
   private val withPreamble: Boolean = true
 ) extends Neo4jQueryStrategy with Logging {
-  private val renderer: CypherRenderer = new CypherRenderer(neo4j)
 
   private val hasSkipLimit: Boolean = partitionPagination.skip != -1 && partitionPagination.topN.limit != -1
 
--- a/src/main/scala/org/neo4j/spark/service/Neo4jQueryService.scala
+++ b/src/main/scala/org/neo4j/spark/service/Neo4jQueryService.scala
@@ -47,7 +47,7 @@ class Neo4jQueryWriteStrategy(
 
   override def createStatementForQuery(options: Neo4jOptions): String = {
     val scriptResult = Neo4jQueryStrategy.scriptResultClause(options)
-    val preamble = if (withPreamble) fullPreamble(neo4j, options.tuning) else ""
+    val preamble = if (withPreamble) fullPreamble(neo4j, options) else ""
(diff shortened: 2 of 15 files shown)


src/test/scala/org/neo4j/spark/service/Neo4jQueryServiceTest.scala
```diff
@@ -1117,147 +1152,113 @@ class Neo4jQueryServiceTest {
...
-    assertEquals(
-      "WITH $scriptResult AS scriptResult MATCH (o:Object) RETURN o",
-      actual
-    )
-  }
-
-  @Test
-  def testScriptResultInclusionInCustomWriteQuery(): Unit = {
+  @ParameterizedTest
+  @MethodSource(Array("versions_prefix_and_tuning_cross_combination"))
```
nord-erik, line 1324: @Emrehzl94 you suggested this, to do testing of all combos of versions and tunings

I thought it would look messy. I'm still a bit over the fence, is this messy or neat?

Running time is still super duper fast, even though we add more than 100+ new cases

Emrehzl94: I don't think it is gonna be slow to run these just now readability is slightly harder. And it brings extra complexity when we want to add new test cases.

Emrehzl94: maybe @fbiville @vasifabdulhamidov-neo can have a better idea about the tests here?

src/main/scala/org/neo4j/spark/cypher/CypherPreamble.scala
```diff
@@ -34,7 +35,15 @@ object CypherPreamble {
...
-  private def versionPreamble(neo4j: Neo4j): String = {
+  private def versionPreamble(neo4j: Neo4j, neo4jOptions: Neo4jOptions): String = {
+    if (neo4jOptions.cypherVersion == "5") {
+      return "CYPHER 5 "
+    }
+
+    if (neo4jOptions.cypherVersion == "25") {
+      return "CYPHER 25 "
+    }
+
```
Emrehzl94: what happens here if my neo4j version doesn't support the cypher version set in options?

nord-erik: Good point, you get a cypher error message:

```
Invalid input '25': expected
  "ALTER"
  "CALL"
  "CREATE"
  "DEALLOCATE"
  "DELETE"
  "DENY"
  "DETACH"
  "DROP"
  "DRYRUN"
  "ENABLE"
  "FOREACH"
  "GRANT"
  "LOAD"
  "MATCH"
  "MERGE"
  "OPTIONAL"
  "REALLOCATE"
  "REMOVE"
  "RENAME"
  "RETURN"
  "REVOKE"
  "SET"
  "SHOW"
  "START"
  "STOP"
  "TERMINATE"
  "UNWIND"
  "USE"
  "USING"
  "WITH" (line 1, column 8 (offset: 7))
"CYPHER 25"
```

if you're older than 5.21 else you get:

```
Neo.ClientError.Statement.ArgumentError
25 is not a valid option for cypher version. Valid options are: 5
```

nord-erik: Do we need to prevent the user from shooting himself in the foot?

src/test/scala/org/neo4j/spark/service/Neo4jQueryServiceTest.scala
nord-erik: agree on your concerns. If we are willing to accept the extra complexity, the test coverage we get from it here is quite nice

src/main/scala/org/neo4j/spark/cypher/CypherPreamble.scala
fbiville: I don't think so. Let the Cypher runtime reject it

Emrehzl94: yeah I think we can do that with a good error message and stuff because it doesn't seem like a big effort, right?

src/test/scala/org/neo4j/spark/service/Neo4jQueryServiceTest.scala
Emrehzl94: I would say let's accept this complexity for now and keep that in mind as improvement for after release if we wanna tackle or feel the need to simplify.

src/main/scala/org/neo4j/spark/cypher/CypherPreamble.scala
```diff
@@ -34,7 +35,15 @@ object CypherPreamble {
     }
   }
 
-  private def versionPreamble(neo4j: Neo4j): String = {
+  private def versionPreamble(neo4j: Neo4j, neo4jOptions: Neo4jOptions): String = {
+    if (neo4jOptions.cypherVersion == "5") {
+      return "CYPHER 5 "
```
fbiville: _thinking out loud_

I like the idea of controlling the preamble completely (vs. just appending whatever the user has specified)
on the other hand, this couples us with Cypher version releases... (low effort, but it's easy to forget... also, we condemn older Spark connector releases to potentially never work with newer Neo4j releases - which is also probably fine)

maybe the threat model here is a bit weak: a Spark user already needs read or write access to the database they're targeting, so I'm not sure letting users write `.option("cypher.option", "5 MATCH (n) CALL (n) {DETACH DELETE n} IN TRANSACTIONS")` would be so bad

src/test/scala/org/neo4j/spark/service/Neo4jQueryServiceTest.scala
```diff
@@ -1117,147 +1150,113 @@ class Neo4jQueryServiceTest {
...
+    mode: String
+  ): Unit = {
     val options: java.util.Map[String, String] = new java.util.HashMap[String, String]()
     options.put(Neo4jOptions.URL, "bolt://localhost")
     options.put(QueryType.QUERY.toString.toLowerCase, "MATCH (o:Object) RETURN o")
+    options.put(Neo4jOptions.CYPHER_VERSION, customVersion)
     val neo4jOptions: Neo4jOptions = new Neo4jOptions(withTuning(options, tuningOptions))
 
     val (strategy, wantQuery) = mode match {
       case "READ" => (
```
fbiville: suggestion: maybe as a follow-up, I'd split READ and WRITE tests (applies to other tests below too)

maybe it's just a personal pet peeve of mine, but I like tests with the lowest cyclomatic complexity as possible ^^

src/test/scala/org/neo4j/spark/service/Neo4jQueryServiceTest.scala
fbiville: as I suggested above, we can remove mode and split tests in two, that would remove at least 1 param
maybe s/customVersion/cypherVersionOption/ and s/versionPrefix/cypherVersionPreamble/ to clarify the parameter names a bit?

src/test/scala/org/neo4j/spark/service/Neo4jQueryServiceTest.scala
nord-erik: Yeah let's do it in follow up PR. I know in this PR I re-wrote a separated case and combined them, but that was for consistency reason. to make all test cases in this snapshot equally strucutred

so future PR will break them all up

src/main/scala/org/neo4j/spark/cypher/CypherPreamble.scala
nord-erik: *assuming* that cypher versions will always be a number, we can check that the string is a numerical string, and pass that. middle ground...

so we pass `s"CHPHER $neo4jOptions.cypherVersion"` and that value is already sanitized/validated

src/main/scala/org/neo4j/spark/cypher/CypherPreamble.scala
nord-erik: I don't think we have any way right now to capture and re-formulate errors from the driver. We just spit out what the driver gives us

src/test/scala/org/neo4j/spark/service/Neo4jQueryServiceTest.scala
nord-erik: I renamed as suggested

I feel like we want to get the coverage in but cases are too dense

I'll start working on the refactor to separate them

src/test/scala/org/neo4j/spark/service/Neo4jQueryServiceTest.scala
nord-erik: I changed my mind, might bake in refactor commit in this PR

src/main/scala/org/neo4j/spark/cypher/CypherPreamble.scala
nord-erik: done

nord-erik: Okay this opened something; we don't validate the options when Spark itself calls `inferSchema` so cypher injection was made possible (use before sanitaztion) if this is carelessly implemented.

Update:

https://github.com/neo4j/neo4j-spark-connector/pull/961

nord-erik:
resolved nit from #963 in last commit to this PR

nord-erik:
resolved concerns of picking CYPHER 25 from 2025.0 onwards, should be 2025.6 onwards.

caniuse predicate will decide if applicable.

src/main/scala/org/neo4j/spark/service/Neo4jQueryService.scala
```diff
@@ -153,14 +152,14 @@ class Neo4jQueryWriteStrategy(
 
 class Neo4jQueryReadStrategy(
   private val neo4j: Neo4j,
+  private val renderer: CypherRenderer,
```
nord-erik, line 155: lifted renderer from internal val to constructor val because by injecting it we can pass Neo4jOptions to it in a more decoupled way

fbiville (approved):
nit: let's not block the PR for that, but it'd be nice to restructure the modified tests (name case, assertions)... let's do it later though

PR #934, opened 2026-05-22
refactor: restructure read tests

fbiville:
This streamlines the usage of TestContainers, relies only on JUnit and Assertj for tests.

This also cleans up a few oddities wrt Scala, Cypher and Spark.


src/test/scala/org/neo4j/spark/ReadIT.scala
```diff
@@ -0,0 +1,1926 @@
...
+    if (spark != null) {
+      spark.close()
+    }
+    if (driver != null) {
+      driver.close()
+    }
+  }
+
+  @Test
+  def throws_exception_if_no_valid_read_option_is_set(): Unit = {
```
fbiville, line 104: note: this test and the test below probably don't need a Neo4j container to cover the behavior we want

src/test/resources/junit-platform.properties
nord-erik, line 1: 🎉

src/test/scala/org/neo4j/spark/ReadIT.scala
```diff
@@ -0,0 +1,1926 @@
...
+        spark.read.format(classOf[DataSource].getName)
+          .option("labels", "Person")
+          .option("relationship", "KNOWS")
+          .load()
+          .show() // show is needed to trigger the exception because of changes in Spark 3
+      })
+      .withMessage("You need to specify just one of these options: 'gds', 'labels', 'query', 'relationship'")
+  }
+
+  @Nested
```
nord-erik, line 127: is it possible to take these nested stuffs and define them in separate files? So we would have ReadIT.scala, ReadByLAbelsIT.scala, ReadBy{QueryRelationshipGds}IT.scala?

fbiville: Yep, I hesitated between the two. It feels to me a tiny bit tidier to limit the number of files, but that means the files are also quite big.

nord-erik:
Let me just screen and see if any of the cherry picks brought in new test cases

src/test/scala/org/neo4j/spark/ReadIT.scala
```diff
@@ -0,0 +1,1927 @@
...
+
+    @Test
+    def returns_selected_localdatetime_column(): Unit = {
+      driver.executableQuery("CREATE (:Person {aTime: localdatetime('2007-12-03T10:15:30')})").execute()
+
+      val df = spark.read.format(classOf[DataSource].getName)
+        .option("labels", "Person")
+        .load()
+
+      val result = df.select("aTime").collectAsList().get(0).getTimestamp(0)
```
nord-erik:
```suggestion
      val result = df.select("aTime").collectAsList().get(0).getAs[LocalDateTime](0)
```

after rebase, because the timestamp change is merged, this assertion needs to assert LocalDateTime

src/test/scala/org/neo4j/spark/ReadIT.scala
```diff
@@ -0,0 +1,1927 @@
...
+      assertThat(result).isEqualTo(Timestamp.from(LocalDateTime.parse("2007-12-03T10:15:30").toInstant(ZoneOffset.UTC)))
```
nord-erik:
```suggestion
      assertThat(result).isEqualTo(LocalDateTime.parse(localDateTime))
```

after rebase, because the timestamp change is merged, this assertion needs to assert LocalDateTime

src/test/scala/org/neo4j/spark/ReadIT.scala
```diff
@@ -0,0 +1,1927 @@
...
+
+      val df = spark.read.format(classOf[DataSource].getName)
+        .option("labels", "Person")
+        .load()
+
+      val result = df.select("scores").collectAsList().get(0).getAs[mutable.Seq[Double]](0)
+      assertThat(result.head).isEqualTo(22.33)
+      assertThat(result(1)).isEqualTo(44.55)
+    }
+
```
nord-erik, line 404: in rebase conflict a test case was added here:

```
  @Test
  def testReadNodeWithTimestampArray(): Unit = {
    val df: DataFrame =
      initTest(
        s"CREATE (p:Person {someTimes: [datetime('2010-10-10T11:13:37+01:00'), datetime('2011-11-11T10:13:37Z')]})"
      )

    val res = df.select("someTimes").collectAsList().get(0).getAs[Seq[Timestamp]](0)
    assertEquals("2010-10-10T10:13:37Z", res.head.toInstant.atZone(ZoneOffset.UTC).toString)
    assertEquals("2011-11-11T10:13:37Z", res(1).toInstant.atZone(ZoneOffset.UTC).toString)
  }
```

src/test/scala/org/neo4j/spark/ReadIT.scala
```diff
@@ -0,0 +1,1927 @@
...
+      assertThat(value1.get(5)).isEqualTo("P0M22DT71509.500000000S")
+      val value2 = row(1)
+      assertThat(value2.get(0)).isEqualTo("duration")
+      assertThat(value2.get(1)).isEqualTo(0L)
+      assertThat(value2.get(2)).isEqualTo(17L)
+      assertThat(value2.get(3)).isEqualTo(43200L)
+      assertThat(value2.get(4)).isEqualTo(0)
+      assertThat(value2.get(5)).isEqualTo("P0M17DT43200S")
+    }
+
```
nord-erik, line 582: In rebase conflict a test case was added here:

```

  @Test
  def testReadNodeWithBinary(): Unit = {
    val bytes = "hello, world!".map(_.toByte).toArray
    val parameters = new java.util.HashMap[String, Object]()
    parameters.put("bytes", bytes)

    SparkConnectorScalaSuiteIT.session()
      .executeWrite((tx: TransactionContext) => tx.run("CREATE (h:Hello {b: $bytes})", parameters).consume())

      .option("labels", "Hello")
      .load()

    val res = df.select("b").collect()
    assertEquals(1, res.length)
    val gotBytes = res.head.getAs[Array[Byte]](0)

    for (i <- bytes.indices) {
      assertEquals(bytes(i), gotBytes(i))
    }
  }
```

src/test/scala/org/neo4j/spark/ReadIT.scala
nord-erik: this was also changed in the APOC test case. Assuming here that via parameter we cover both using this one test case!

nord-erik (approved):

PR #963, opened 2026-06-01
feat: major bump cypher DSL

nord-erik:
this PR bumps cypher DSL and updates breaking changes

should yield no regressions in our test cases

### DIFF
--- a/src/main/scala/org/neo4j/spark/cypher/CypherRenderer.scala
+++ b/src/main/scala/org/neo4j/spark/cypher/CypherRenderer.scala
@@ -37,7 +37,7 @@ case class CypherRenderer(neo4j: Neo4j) {
 
   private def getDialect(neo4j: Neo4j): Dialect = {
     if (neo4j.getVersion.compareTo(Neo4jV5) < 0) {
-      throw new IllegalArgumentException("Neo4j version 5.0 or higher is required to use this connector version")
+      throw new IllegalArgumentException("Neo4j version 5.26 or higher is required to use this connector version")
     }
 
     if (neo4j.getVersion.compareTo(Neo4jV5_23) < 0) {
--- a/src/main/scala/org/neo4j/spark/cypher/CypherRenderer.scala
+++ b/src/main/scala/org/neo4j/spark/cypher/CypherRenderer.scala
@@ -21,29 +21,38 @@ import org.neo4j.caniuse.Neo4jVersion
 import org.neo4j.cypherdsl.core.Statement
 import org.neo4j.cypherdsl.core.renderer.Configuration
 import org.neo4j.cypherdsl.core.renderer.Dialect
+import org.neo4j.cypherdsl.core.renderer.GeneralizedRenderer
 import org.neo4j.cypherdsl.core.renderer.Renderer
 import org.neo4j.spark.cypher.CypherRenderer.Neo4jV5
+import org.neo4j.spark.cypher.CypherRenderer.Neo4jV5_23
 
-class CypherRenderer(neo4j: Neo4j) extends Renderer {
+case class CypherRenderer(neo4j: Neo4j) {
 
   private val cached =
     Renderer.getRenderer(
       Configuration.newConfig()
-        .withDialect(
-          if (neo4j.getVersion.compareTo(Neo4jV5) < 0) {
-            Dialect.DEFAULT
-          } else {
-            Dialect.NEO4J_5
-          }
-        )
+        .withDialect(getDialect(neo4j))
         .build()
     )
 
-  override def render(statement: Statement): String = {
+  private def getDialect(neo4j: Neo4j): Dialect = {
+    if (neo4j.getVersion.compareTo(Neo4jV5) < 0) {
+      throw new IllegalArgumentException("Neo4j version 5.0 or higher is required to use this connector version")
+    }
+
+    if (neo4j.getVersion.compareTo(Neo4jV5_23) < 0) {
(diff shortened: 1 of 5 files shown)


src/main/scala/org/neo4j/spark/cypher/CypherRenderer.scala
```diff
@@ -21,29 +21,38 @@ import org.neo4j.caniuse.Neo4jVersion
...
-          }
-        )
+        .withDialect(getDialect(neo4j))
         .build()
     )
 
-  override def render(statement: Statement): String = {
+  private def getDialect(neo4j: Neo4j): Dialect = {
+    if (neo4j.getVersion.compareTo(Neo4jV5) < 0) {
+      throw new IllegalArgumentException("Neo4j version 5.0 or higher is required to use this connector version")
```
fbiville:
```suggestion
      throw new IllegalArgumentException("Neo4j version 5.26 or higher is required to use this connector version")
```

fbiville: maybe let's update the logic of that whole method to assume that version >= 5.26?

fbiville: either now or as a follow-up PR

nord-erik: Isn't 5.0 supported for 6.0 connector? Are we cutting form 5.26 onwards?

nord-erik: Ah the test matrix just say "5" and I've been assuming 5.0+, but ofc we should support LTS onwards

neo4j-connectors:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>Commit Message 'Apply suggestions from code review

Co-authored-by: Florent Biville <445792+fbiville@users.noreply.github.com>': subject may not be empty, type may not be empty</td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 0d63dad27564447f84c416735c85785732778f63
</p>

src/main/scala/org/neo4j/spark/cypher/CypherRenderer.scala
```diff
@@ -17,33 +17,22 @@
...
-import org.neo4j.caniuse.Neo4jVersion
 import org.neo4j.cypherdsl.core.Statement
 import org.neo4j.cypherdsl.core.renderer.Configuration
 import org.neo4j.cypherdsl.core.renderer.Dialect
+import org.neo4j.cypherdsl.core.renderer.GeneralizedRenderer
 import org.neo4j.cypherdsl.core.renderer.Renderer
-import org.neo4j.spark.cypher.CypherRenderer.Neo4jV5
 
-class CypherRenderer(neo4j: Neo4j) extends Renderer {
+case class CypherRenderer(neo4j: Neo4j) {
```
ali-ince: do we use `neo4j` inside this case class anywhere?

src/main/scala/org/neo4j/spark/DataSource.scala
```diff
@@ -35,6 +36,8 @@ class DataSource extends TableProvider
 
   Validations.validate(ValidateSparkMinVersion("4.0.0"))
 
+  private val minimumAllowedNeo4jVersion = new Neo4jVersion(5, 26, 0)
```
ali-ince: Are we sure that we want to force 5.26 as the minimum version? I'm not 100% sure about the distribution but I expect we'll see some users on earlier versions of 5.

src/main/scala/org/neo4j/spark/cypher/CypherRenderer.scala
nord-erik: No sorry latest revision stopped using it good call

src/main/scala/org/neo4j/spark/DataSource.scala
nord-erik: removed any type of hard check, but dialect picks based on best effort within 5.x

src/main/scala/org/neo4j/spark/cypher/CypherRenderer.scala
```diff
@@ -22,28 +22,31 @@ import org.neo4j.cypherdsl.core.Statement
...
-        .build()
-    )
-
-  override def render(statement: Statement): String = {
+import org.neo4j.spark.cypher.CypherRenderer.Neo4jV5_23
+import org.neo4j.spark.cypher.CypherRenderer.Neo4j_2025
+
+case class CypherRenderer(private val neo4j: Neo4j) {
+
+  private def dialect(): Dialect = {
```
fbiville, line 30: nit: caller-before-called ordering would move `dialect` below `cached` (and public-before-private would make `dialect` go below `render`) 🤓

fbiville (approved):

src/main/scala/org/neo4j/spark/cypher/CypherRenderer.scala
nord-erik: will fix will fix

nord-erik: I will fix this is in the future, I have already local commits on top of this PR for future stuffs

PR #970, opened 2026-06-02
feat: move streaming tests to separate test file

vasifabdulhamidov-neo:
Reorg tests: Moving streaming tests into separate StreamingIT file following new idea from ReadIT test structures


src/test/scala/org/neo4j/spark/StreamingIT.scala
```diff
@@ -0,0 +1,650 @@
...
+
+import scala.collection.immutable
+import scala.collection.mutable
+import scala.jdk.CollectionConverters.SeqHasAsJava
+
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+@DisplayName("streaming")
+class StreamingIT {
+
+  private val neo4jContainer: Neo4jContainer = new Neo4jContainer(TestUtil.neo4jImage())
```
fbiville: suggestion: I'd rely on `Neo4jContainerProvider` here, just like in `ReadIT`, so that we have streaming tests both for vanilla Neo4j and for Neo4j with the APOC plugin

src/test/scala/org/neo4j/spark/StreamingIT.scala
```diff
@@ -0,0 +1,650 @@
...
+    Option(query).foreach(_.stop())
+    Option(spark).foreach { session =>
+      createdTables.foreach(table => session.sql(s"DROP TABLE IF EXISTS $table"))
+      createdTables.clear()
+      session.close()
+    }
+    Option(driver).foreach(_.close())
+  }
+
+  private def checkpoint(): String =
```
fbiville: nit: I'd move `private` functions to the bottom

src/test/scala/org/neo4j/spark/StreamingIT.scala
```diff
@@ -0,0 +1,650 @@
...
+  }
+
+  @Nested
+  @DisplayName("from source")
+  class FromSource {
+
+    private val total = 60
+
+    @Test
+    @DisplayName("reads nodes created after the stream started")
```
fbiville: suggestion: you can drop `@DisplayName` since we have a custom display name generator now (see https://github.com/neo4j/neo4j-spark-connector/blob/23f08c1b20e8c798bc7e33fd9900866d5fc6e8ed/src/test/resources/junit-platform.properties#L1)
it basically replaces `_` from the test name to ` `

src/test/scala/org/neo4j/spark/StreamingIT.scala
```diff
@@ -0,0 +1,650 @@
...
+        .option(OptFrom, "NOW")
+        .load()
+        .writeStream
+        .format("memory")
+        .queryName("nodesFromNow")
+        .start()
+
+      Executors.newSingleThreadExecutor().execute(() => createMovieNodes(1, total, 1000, 200))
+
+      val expected: immutable.Seq[Map[String, Any]] = (1 to total)
```
fbiville: nit
```suggestion
      val expected = (1 to total)
```
applies below too

src/test/scala/org/neo4j/spark/StreamingIT.scala
```diff
@@ -0,0 +1,650 @@
...
+      "source.age" -> index,
+      "<target.labels>" -> List("Post"),
+      "target.hash" -> s"hash$index",
+      "rel.id" -> index
+    )
+
+    private def selectMovies(sql: String): immutable.Seq[Map[String, Any]] =
+      spark.sql(sql).collect().map { row =>
+        Map(
+          "<labels>" -> row.getAs[Seq[String]]("<labels>"),
```
fbiville: nit: this could be a tiny bit more generic, by accepting an extra parameter: `Map<Class<?>, String>` (with e.g. `Map(classOf[Seq[String]] -> "<labels>", classOf[String] -> "title")` at the callsite)
with that change, you'd only need 1 select method

fbiville (approved):

PR #972, opened 2026-06-03
refactor: use cypher dsl on read/write queries we generate

nord-erik:
refactor of Neo4jQueryService to use cypher dsl library instead of
concatenating strings

one side effect of this is that cypher dsl has a good tendency to
escape identifiers so some queries are different

finally, the dsl prefer spaces over new-line so test cases are
adjusted to replace new line assertions with space assertions.

### DIFF
--- a/src/main/scala/org/neo4j/spark/service/Neo4jQueryService.scala
+++ b/src/main/scala/org/neo4j/spark/service/Neo4jQueryService.scala
@@ -29,10 +29,12 @@ import org.neo4j.caniuse.Neo4j
 import org.neo4j.cypherdsl.core._
 import org.neo4j.spark.cypher.CypherPreamble.fullPreamble
 import org.neo4j.spark.cypher.CypherRenderer
-import org.neo4j.spark.service.Neo4jQueryStrategy.unwindEventsClause
+import org.neo4j.spark.service.Neo4jQueryStrategy.VARIABLE_EVENT
+import org.neo4j.spark.service.Neo4jQueryStrategy.unwindEventsAsEvent
 import org.neo4j.spark.util.Neo4jImplicits._
 import org.neo4j.spark.util.Neo4jNodeMetadata
 import org.neo4j.spark.util.Neo4jOptions
+import org.neo4j.spark.util.Neo4jRelationshipMetadata
 import org.neo4j.spark.util.Neo4jTuningOptions
 import org.neo4j.spark.util.Neo4jUtil
 import org.neo4j.spark.util.NodeSaveMode
@@ -50,26 +52,7 @@ class Neo4jQueryWriteStrategy(
   override def createStatementForQuery(options: Neo4jOptions): String = {
     val preamble = if (withPreamble) fullPreamble(neo4j, options) else ""
     val scriptResult = Neo4jQueryStrategy.scriptResultClause(options)
-    val prefixedUnwind = preamble + scriptResult + unwindEventsClause
-    prefixedUnwind + options.query.value
-  }
-
-  private def cypherNode(nodeData: Neo4jNodeMetadata, alias: String): Node = {
-    val n = Cypher.node(nodeData.labels.head, nodeData.labels.tail: _*).named(alias)
-
-    val keyProperties =
-      nodeData.nodeKeys.toSeq.flatMap { case (_, keyName) =>
-        Seq(
-          keyName,
-          Cypher.property(Cypher.name(Neo4jQueryStrategy.VARIABLE_EVENT), Neo4jWriteMappingStrategy.KEYS, keyName)
-        )
-      }
-
-    if (keyProperties.nonEmpty) {
-      n.withProperties(keyProperties: _*)
-    } else {
-      n
-    }
+    preamble + scriptResult + unwindEventsAsEvent + options.query.value
   }
 
   private def createPropsList(props: Map[String, String], prefix: String): String = {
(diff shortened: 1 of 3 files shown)


src/main/scala/org/neo4j/spark/service/Neo4jQueryService.scala
```diff
@@ -29,121 +29,159 @@ import org.neo4j.caniuse.Neo4j
...
+            propertyPrefix,
+            Neo4jWriteMappingStrategy.KEYS,
+            from
+          )
+        }
+      )
+    }
+  }
+
+  private def cypherRelationshipKeys(mappings: Seq[(String, String)]): Seq[Object] = {
```
nord-erik, line 174: I think this is a bug, because the only difference with `cypherNodeKeys` is that for nodes we do not apply the mapping but for relationships we do.

Still porting this difference as-is to keep the Cypher-DSL re-factor clean

nord-erik:
Bug spotted after refactor, we never throw ErrorIfExist validation error:

https://github.com/neo4j/neo4j-spark-connector/pull/979

src/main/scala/org/neo4j/spark/service/Neo4jQueryService.scala
```diff
@@ -29,121 +29,161 @@ import org.neo4j.caniuse.Neo4j
...
+        case NodeSaveMode.Append    => sourceWith.create(targetNode).mutate(targetNode, targetPropsName)
+        case NodeSaveMode.Match     => sourceWith.`match`(targetNode)
+      }
     } else {
-      ""
+      options.relationshipMetadata.targetSaveMode match {
+        case NodeSaveMode.Overwrite              => sourceMatcher.merge(targetNode).mutate(targetNode, targetPropsName)
+        case NodeSaveMode.Append                 => sourceMatcher.create(targetNode).mutate(targetNode, targetPropsName)
+        case NodeSaveMode.Match if isSourceMatch => Cypher.`match`(sourceNode, targetNode)
+        case _ => throw new IllegalStateException("Impossible state reached. Please report this as a bug.")
```
nord-erik: technically this match statement is not exhaustive (case when `NodeSaveMode.Match if !isSourceMatch`) but we should never ever reach that state because of the outer if statement 🤓

src/main/scala/org/neo4j/spark/service/Neo4jQueryService.scala
```diff
@@ -29,121 +29,160 @@ import org.neo4j.caniuse.Neo4j
...
+import org.neo4j.spark.util.Neo4jUtil.RELATIONSHIP_SOURCE_ALIAS
+import org.neo4j.spark.util.Neo4jUtil.RELATIONSHIP_TARGET_ALIAS
 import org.neo4j.spark.util.NodeSaveMode
 import org.neo4j.spark.util.QueryType
 
 import scala.jdk.CollectionConverters.SeqHasAsJava
 
 class Neo4jQueryWriteStrategy(
   private val neo4j: Neo4j,
+  private val renderer: CypherRenderer,
```
Emrehzl94, line 54: why did we want to pass renderer as an argument to the class? couldn't we just create it in the constructor?

src/main/scala/org/neo4j/spark/service/Neo4jQueryService.scala
```diff
@@ -29,121 +29,160 @@ import org.neo4j.caniuse.Neo4j
...
-      .map(_.quote())
-      .mkString(":")
+    val preamble = if (withPreamble) fullPreamble(neo4j, options) else ""
+    val query = renderer.render(nodeMatcher.mutate(node, eventProperties).build())
 
-    val keys = createPropsList(
-      options.nodeMetadata.nodeKeys,
-      Neo4jWriteMappingStrategy.KEYS
-    )
+    preamble + Neo4jQueryStrategy.unwindEventsAsEvent + query
```
Emrehzl94:
```suggestion
    preamble + unwindEventsAsEvent + query
```

src/main/scala/org/neo4j/spark/service/Neo4jQueryService.scala
nord-erik: https://github.com/neo4j/neo4j-spark-connector/pull/933#discussion_r3349567654

> lifted renderer from internal val to constructor val because by injecting it we can pass Neo4jOptions to it in a more decoupled way

In principle this makes CypherRenderer easier to test. But ironically we do not have any tests for it (yet).

src/main/scala/org/neo4j/spark/service/Neo4jQueryService.scala
```diff
@@ -29,121 +29,160 @@ import org.neo4j.caniuse.Neo4j
...
+  }
+
+  private def cypherNode(nodeData: Neo4jNodeMetadata, alias: String, prefix: Boolean = false): Node = {
```
Emrehzl94, line 128: do we need prefix argument here? can we not just check if alias is blank or not?

src/main/scala/org/neo4j/spark/service/Neo4jQueryService.scala
Emrehzl94: I would say let's remove it then to prevent further confusions or throw an exception with a message indicating the case?

src/main/scala/org/neo4j/spark/service/Neo4jQueryService.scala
nord-erik: we want prefix for relationship.source and relationship.target when generating pattern for a relationship query. However we don't want the prefix if we're generating pattern for node query

so the logic is sadly not as simple as checking if alias is empty (it never is)

we could check if alias == node, but that creates a pretty quirky side effect

for sure this is a tech-debt from old implementation, maybe there's a way to refactor it into something cleaner

src/main/scala/org/neo4j/spark/service/Neo4jQueryService.scala
nord-erik: Agree

Emrehzl94 (approved):

PR #979, opened 2026-06-08
fix: validations for ErrorIfExists

nord-erik:
Logic for when to disallow ErrorIfExists was wrong, and the test case
covering it did not take into account for when no exception was
thrown.

From what I could tell, this exception was never thrown and the test
case was false positive.

Fixed this by updating the test to also assert that a test case was
indeed thrown. Then fixed the logic for throwing; if any save mode
is 'ErrorIfExists' then we should throw.

### DIFF
--- a/src/main/scala/org/neo4j/spark/writer/Neo4jWriterBuilder.scala
+++ b/src/main/scala/org/neo4j/spark/writer/Neo4jWriterBuilder.scala
@@ -56,7 +56,7 @@ class Neo4jWriterBuilder(
         ValidationUtil.isFalse(
           o.relationshipMetadata.sourceSaveMode.equals(NodeSaveMode.ErrorIfExists)
             || o.relationshipMetadata.targetSaveMode.equals(NodeSaveMode.ErrorIfExists),
-          "Save mode 'ErrorIfExists' is not supported on Spark 3.0, use 'Append' instead."
+          "This connector does not support save mode 'ErrorIfExists'. Use save mode 'Append' instead."
         )
       }
     ))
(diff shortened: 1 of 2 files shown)


src/main/scala/org/neo4j/spark/writer/Neo4jWriterBuilder.scala
```diff
@@ -55,7 +55,7 @@ class Neo4jWriterBuilder(
       (o: Neo4jOptions) => {
         ValidationUtil.isFalse(
           o.relationshipMetadata.sourceSaveMode.equals(NodeSaveMode.ErrorIfExists)
-            && o.relationshipMetadata.targetSaveMode.equals(NodeSaveMode.ErrorIfExists),
+            || o.relationshipMetadata.targetSaveMode.equals(NodeSaveMode.ErrorIfExists),
           "Save mode 'ErrorIfExists' is not supported on Spark 3.0, use 'Append' instead."
```
Emrehzl94: do we need to change the message here?

Emrehzl94: Looks like ErrorIfExists is supported since spark 1.4.0 maybe the message should refer to connector 4.0 instead of spark.

fbiville: Spark v3 started using DataSource v2 APIs where our implementation of table scans started to interfere with `ErrorIfExists`
The message could however be clearer: `This connector does not support save mode 'ErrorIfExists' since the release of Spark core 3.0, use 'Append' instead`

nord-erik: Left out any spark core versioning notation because it's irrelevant. For all spark versions we support, no version enables ErrorIfExists feature in our connector.

Emrehzl94 (approved):

PR #980, opened 2026-06-08
fix: validations for ErrorIfExists

nord-erik:
Logic for when to disallow ErrorIfExists was wrong, and the test case
covering it did not take into account for when no exception was
thrown.

From what I could tell, this exception was never thrown and the test
case was false positive.

Fixed this by updating the test to also assert that a test case was
indeed thrown. Then fixed the logic for throwing; if any save mode
is 'ErrorIfExists' then we should throw.


nord-erik:
let's re-visit if we need this for 5.0 later on but atm I don't think so

PR #984, opened 2026-06-12
feat: cypher parameters are now fully dynamic

nord-erik:
Having white-list/static style of allowed cypher tuning parameters in
the preamble will likely not scale in the future if new parameters are
added in the cypher lang.

In hindsight it is much better to allow dynamic parameters for the
preamble. This PR introduces dynamic options for cypher tunings.

The options prefix is: "cypher.tuning.". Every option you set with this
prefix will then be passed to the `CYPHER` preamble as key-value pairs.

Examples:

```
.options("cypher.tuning.runtime", "parallel") =>
CYPHER runtime=parallel

.options("cypher.tuning.anyParameterYouWish", "set_to_value") =>
CYPHER anyParameterYouWish=set_to_value
```

Only validations run is that the parameter name must be alphanumeric
and must not contain any spaces or dots. However hyphens and
underscores are allowed.

### DIFF
--- a/src/main/scala/org/neo4j/spark/cypher/CypherPreamble.scala
+++ b/src/main/scala/org/neo4j/spark/cypher/CypherPreamble.scala
@@ -20,13 +20,12 @@ import org.neo4j.caniuse.CanIUse.INSTANCE.canIUse
 import org.neo4j.caniuse.Cypher.{INSTANCE => Cypher}
 import org.neo4j.caniuse.Neo4j
 import org.neo4j.spark.util.Neo4jOptions
-import org.neo4j.spark.util.Neo4jTuningOptions
 
 object CypherPreamble {
 
   def fullPreamble(neo4j: Neo4j, neo4jOptions: Neo4jOptions): String = {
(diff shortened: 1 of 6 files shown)


src/main/scala/org/neo4j/spark/cypher/CypherPreamble.scala
```diff
@@ -67,4 +62,29 @@ object CypherPreamble {
...
+      throw new IllegalArgumentException(
+        "Cypher tuning parameter value must be alphanumeric, underscore or hyphen. Found: " + parameterValue
+      )
+    }
+
+    true
+  }
+
+  private def toCypherParameterName(parameter: String): String = {
+    parameter.split('.').zipWithIndex.map {
```
fbiville: question: is there a hard rule preventing cypher options to be dotted?
I'm not sure we need that translation logic or whether that would create friction in the future

nord-erik: lets toss it

fbiville (approved):

PR #989, opened 2026-06-16
feat: wrap user-defined read query as CALL subquery

fbiville:
This leverages Cypher DSL's parser and callback-based API to alias
returned columns, if necessary, before the user-defined query gets
wrapped into a containing `CALL` clause.

```cypher
MATCH (n) RETURN n.name
```

**cannot** be transformed into:

```cypher
CALL () { MATCH (n) RETURN n.name } RETURN * // invalid!
// error: Expression in CALL { RETURN ... } must be aliased (use AS)
```

as `n` is not available outside the inner scope.

The new `AutomaticAliaser` facility turns the initial query into:

```cypher
MATCH (n) RETURN n.name AS `n.name`
```

That way, most user-defined queries can be embedded as CALL
subqueries.
As a consequence, previous restrictions such as the rejection
of queries with `SKIP` or `LIMIT` do not apply anymore.

One restriction that still applies is that simple procedure calls
(without `YIELD`) will not work. Subqueries must use `RETURN`.

This commit also now fully applies TopN optimizations to
user-defined queries.

### DIFF
--- a/src/test/scala/org/neo4j/spark/cypher/AutomaticAliaserIT.scala
+++ b/src/test/scala/org/neo4j/spark/cypher/AutomaticAliaserIT.scala
@@ -34,6 +34,7 @@ import org.neo4j.cypherdsl.core.Cypher.asterisk
 import org.neo4j.cypherdsl.core.Cypher.callRawCypher
 import org.neo4j.driver.Driver
 import org.neo4j.driver.exceptions.ClientException
+import org.neo4j.spark.testsupport.Neo4jContainerProvider.ADMIN_PASSWORD
 import org.neo4j.spark.testsupport.Neo4jExtensions.Neo4jContainerExtensions
 import org.neo4j.spark.testsupport.TestUtil
 import org.testcontainers.junit.jupiter.Container
@@ -50,7 +51,7 @@ class AutomaticAliaserIT {
 
   @Container
   private val container = new Neo4jContainer(TestUtil.neo4jImage())
-    .withAdminPassword("letmein!")
+    .withAdminPassword(ADMIN_PASSWORD)
     .withEnv("NEO4J_ACCEPT_LICENSE_AGREEMENT", "yes")
 
   private var driver: Option[Driver] = None
--- a/src/test/scala/org/neo4j/spark/cypher/AutomaticAliaserIT.scala
+++ b/src/test/scala/org/neo4j/spark/cypher/AutomaticAliaserIT.scala
@@ -1,0 +1,169 @@
+/*
+ * Copyright (c) "Neo4j"
+ * Neo4j Sweden AB [https://neo4j.com]
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.neo4j.spark.cypher
+
+import org.assertj.core.api.Assertions.assertThat
+import org.assertj.core.api.Assertions.assertThatCode
+import org.assertj.core.api.Assertions.assertThatObject
+import org.assertj.core.api.Assertions.assertThatThrownBy
+import org.junit.jupiter.api.AfterEach
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.TestInstance
+import org.junit.jupiter.api.TestInstance.Lifecycle.PER_CLASS
+import org.junit.jupiter.params.ParameterizedTest
+import org.junit.jupiter.params.provider.Arguments
+import org.junit.jupiter.params.provider.Arguments.argumentSet
+import org.junit.jupiter.params.provider.MethodSource
+import org.neo4j.cypherdsl.core.Cypher
+import org.neo4j.cypherdsl.core.Cypher.asterisk
+import org.neo4j.cypherdsl.core.Cypher.callRawCypher
+import org.neo4j.driver.Driver
+import org.neo4j.driver.exceptions.ClientException
+import org.neo4j.spark.testsupport.Neo4jExtensions.Neo4jContainerExtensions
+import org.neo4j.spark.testsupport.TestUtil
+import org.testcontainers.junit.jupiter.Container
+import org.testcontainers.junit.jupiter.Testcontainers
+import org.testcontainers.neo4j.Neo4jContainer
+
+import java.util.stream.Stream
+
+import scala.jdk.CollectionConverters.MapHasAsJava
+
+@Testcontainers
+@TestInstance(PER_CLASS)
+class AutomaticAliaserIT {
+
+  @Container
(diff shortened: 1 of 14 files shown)


src/test/scala/org/neo4j/spark/cypher/AutomaticAliaserTest.scala
```diff
@@ -0,0 +1,169 @@
...
+      ),
+      argumentSet(
+        "unaliased string literal",
+        "RETURN 'foo'",
+        "RETURN 'foo' AS `'foo'`"
+      ),
+      argumentSet(
+        "unaliased array literal",
+        "RETURN ['foo']",
+        "RETURN ['foo'] AS `['foo']`"
```
nord-erik, line 69: am I reading this right,

does this mean that the return record will have literally, key >>`['foo']`<< with brackets and all?

what if the list is very long? does this end up as a column name in case of reads?

nord-erik: but yeah I guess recommended use case is to use alias in the custom query

src/test/scala/org/neo4j/spark/cypher/AutomaticAliaserIT.scala
```diff
@@ -0,0 +1,169 @@
...
+
+import scala.jdk.CollectionConverters.MapHasAsJava
+
+@Testcontainers
+@TestInstance(PER_CLASS)
+class AutomaticAliaserIT {
+
+  @Container
+  private val container = new Neo4jContainer(TestUtil.neo4jImage())
+    .withAdminPassword("letmein!")
```
nord-erik: maybe we can re-use constant `Neo4jContainerProvider.ADMIN_PASSWORD` or can we even use `Neo4jContainerProvider`?

src/test/scala/org/neo4j/spark/cypher/AutomaticAliaserTest.scala
fbiville: yep, I think it's easier to literally take the column name and escape it

src/test/scala/org/neo4j/spark/cypher/AutomaticAliaserIT.scala
```diff
@@ -0,0 +1,169 @@
...
+
+  @AfterEach
+  def cleanUp(): Unit = {
+    driver.foreach(_.close())
+  }
+
+  @ParameterizedTest
+  @MethodSource(Array("alias_cases"))
+  def aliases_query_for_subquery_embedding(subquery: String, params: Map[String, AnyRef]): Unit = {
+    assertThat(driver.isDefined).isTrue
```
nord-erik, line 76: still need this?

fbiville: it's just to self-document that unwrapping the driver `Option` after that is safe

src/test/scala/org/neo4j/spark/cypher/AutomaticAliaserTest.scala
fbiville: it's either that or we process the column name by stripping some characters... which is, in my opinion, harder to reason about for users

src/test/scala/org/neo4j/spark/cypher/AutomaticAliaserIT.scala
fbiville: I'm on the fence on `Neo4jContainerProvider`. This integration test only really exercises Cypher and does not even use Spark. I'm not sure if covering various Neo4j distributions is worth it at that level.

fbiville: I can make the password field public and reuse it here.

fbiville: done in 21e40318bc0f5931e761a66bf3e89e9f0499d8df

src/test/scala/org/neo4j/spark/cypher/AutomaticAliaserTest.scala
nord-erik: I think this is fine, might create some oopsies but indeed we can't assume what else to put as name

src/test/scala/org/neo4j/spark/cypher/AutomaticAliaserIT.scala
nord-erik: nice!

nord-erik (approved):
looks great, nice work!

src/test/scala/org/neo4j/spark/service/Neo4jQueryServiceTest.scala
```diff
@@ -1108,7 +1128,12 @@ class Neo4jQueryServiceTest {
         )
       ).createQuery()
 
-      assertThat(query).isEqualTo(s"${prefix}MATCH (p:Person) RETURN p SKIP 0 LIMIT 24")
+      val callParens = if (neo4j.getVersion.compareTo(new Neo4jVersion(5, 23, 0)) >= 0) " () " else " "
```
ali-ince: nit: this could be replaced by `Cypher.callSubqueryWithVariableScopeClause` from caniuse library.

src/test/scala/org/neo4j/spark/service/Neo4jQueryServiceTest.scala
```diff
@@ -1196,7 +1221,9 @@ class Neo4jQueryServiceTest {
       val neo4jOptions = new Neo4jOptions(optionsMap ++ tuningOptions)
       val readService = new Neo4jQueryService(neo4jOptions, plainReadStrategy(neo4j, neo4jOptions))
 
-      val wantQuery = s"$tuningPrefix\n${cypherVersionPreamble}MATCH (o:Object) RETURN o"
+      val callParens = if (neo4j.getVersion.compareTo(new Neo4jVersion(5, 23, 0)) >= 0) " () " else " "
```
ali-ince: same here.

src/test/scala/org/neo4j/spark/service/Neo4jQueryServiceTest.scala
```diff
@@ -1219,14 +1246,15 @@ class Neo4jQueryServiceTest {
       val neo4jOptions = new Neo4jOptions(optionsMap ++ tuningOptions)
       val versionedReadService = new Neo4jQueryService(neo4jOptions, plainReadStrategy(neo4j, neo4jOptions))
 
+      val callParens = if (neo4j.getVersion.compareTo(new Neo4jVersion(5, 23, 0)) >= 0) " () " else " "
```
ali-ince: same here.

ali-ince (approved):
LGTM, Just had one nitpick.

src/test/scala/org/neo4j/spark/service/Neo4jQueryServiceTest.scala
fbiville: done in 9fefb580579ddb802566b269083afff8df745c8c

src/test/scala/org/neo4j/spark/service/Neo4jQueryServiceTest.scala
fbiville: done in 9fefb580579ddb802566b269083afff8df745c8c

src/test/scala/org/neo4j/spark/service/Neo4jQueryServiceTest.scala
fbiville: done in 9fefb580579ddb802566b269083afff8df745c8c

PR #991, opened 2026-06-16
refactor: use scala map for Neo4jOptions

nord-erik:
Swaps from java.util.Map to Map in `Neo4jOptions`.

Reasoning that test cases should be easier to write with declarative
creating maps rather that procedurally creating maps.

The past way of doing it created lots of cases where we wrote Java-like
code in our Scala bits. Ultimately aesthetic refactor. However we do
see simpler test case setups imo

### DIFF
--- a/src/test/scala/org/neo4j/spark/cypher/CypherPreambleTest.scala
+++ b/src/test/scala/org/neo4j/spark/cypher/CypherPreambleTest.scala
@@ -57,14 +57,9 @@ class CypherPreambleTest {
       dynamicTest(
         "when " + testCase.name,
         () => {
-          val options = new java.util.HashMap[String, String]()
-          options.put(Neo4jOptions.URL, "stub-url-doesnt-matter-but-is-required")
+          val neo4jOptions =
+            new Neo4jOptions(testCase.tuningOptions + (Neo4jOptions.URL -> "stub-url-doesnt-matter-but-is-required"))
 
-          testCase.tuningOptions.foreach { case (key, value) =>
-            options.put(key, value)
-          }
-
-          val neo4jOptions = new Neo4jOptions(options.asScala.toMap)
           val tuningPreamble = CypherPreamble.tuningPreamble(neo4jOptions)
 
           assertThat(tuningPreamble).isEmpty()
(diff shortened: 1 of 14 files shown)


src/test/scala/org/neo4j/spark/cypher/CypherPreambleTest.scala
```diff
@@ -109,17 +105,14 @@ class CypherPreambleTest {
...
-
-          testCase.tuningOptions.foreach { case (key, value) =>
-            options.put(key, value)
-          }
+          val neo4jOptions =
+            new Neo4jOptions(testCase.tuningOptions + (Neo4jOptions.URL -> "stub-url-doesnt-matter-but-is-required"))
 
-          val neo4jOptions = new Neo4jOptions(options)
-          val tuningPreamble = CypherPreamble.tuningPreamble(neo4jOptions)
+          // we don't care about the order of peramble params, but we care about their individual shape
```
nord-erik, line 111: this started failing when changing the underlying map. My best guess is simple that ordering is not guaranteed for maps. So removed order-sensitivity here.

fbiville (approved):

src/test/scala/org/neo4j/spark/cypher/CypherPreambleTest.scala
fbiville: if Scala `Map` defaults to `HashMap`, then there's no guaranteed ordering indeed
if you want insertion order -> `LinkedHashMap`
if you want sorting -> `TreeMap`

fbiville: I tend to use `LinkedHashMap` a lot, since this makes testing a lot easier

nord-erik: Thanks good idea didn't think of passing non-standerd Map. I will change this because the old assertion was much clearer

nord-erik: I'll skip doing this in the end because Neo4jOptions expect a standard map so the test case became a mess

PR #992, opened 2026-06-16
feat: support indexed script options

Emrehzl94:
This PR adds support for scripts as individual indexed options (script.1, script.2, ...) as an alternative to the semicolon delimited script string.


vasifabdulhamidov-neo (approved):
lgtm

src/main/scala/org/neo4j/spark/util/Neo4jOptions.scala
```diff
@@ -365,6 +362,40 @@ class Neo4jOptions(private val options: java.util.Map[String, String]) extends S
       .map(t => (t._1.substring(CYPHER_TUNING_OPTION_PREFIX.length), t._2))
       .toMap
 
+  private def extractScript(): Array[String] = {
+    val fromScript = getParameter(SCRIPT)
+      .split(";")
```
ali-ince: I think we decided to remove this split logic from the `script` option.

src/main/scala/org/neo4j/spark/util/Neo4jOptions.scala
```diff
@@ -365,6 +362,40 @@ class Neo4jOptions(private val options: java.util.Map[String, String]) extends S
...
+      .map(_.trim)
+      .filterNot(_.isEmpty)
+
+    val fromScriptOptions = options.asScala
+      .view
+      .filterKeys(_.startsWith(SCRIPT_PREFIX))
+      .map { case (key, value) =>
+        val index = key.substring(SCRIPT_PREFIX.length)
+        if (!index.matches("[1-9]\\d*")) {
```
ali-ince: how about just `\d+` so that they could also use `script.0`, `script.1`, etc.?

src/main/scala/org/neo4j/spark/util/Neo4jOptions.scala
```diff
@@ -365,6 +362,40 @@ class Neo4jOptions(private val options: java.util.Map[String, String]) extends S
...
+          throw new IllegalArgumentException(
+            s"Script option '$key' must use a positive integer suffix"
```
ali-ince:
```suggestion
            s"Script option '$key' must have an integer suffix"
```

src/main/scala/org/neo4j/spark/util/Validations.scala
```diff
@@ -261,7 +261,7 @@ case class ValidateWrite(
       neo4jOptions.script.foreach(query =>
         ValidationUtil.isTrue(
           schemaService.isValidQuery(query, new java.util.HashMap[String, AnyRef]()),
-          s"The following query inside the `${Neo4jOptions.SCRIPT}` is not valid, please check the syntax: $query"
+          s"The following query inside the script options (`${Neo4jOptions.SCRIPT}` / `${Neo4jOptions.SCRIPT_PREFIX}N`) is not valid, please check the syntax: $query"
```
ali-ince: Maybe just


```suggestion
          s"The following script query is not valid, please check the syntax: $query"
```

src/main/scala/org/neo4j/spark/util/Validations.scala
```diff
@@ -349,7 +349,7 @@ case class ValidateRead(neo4j: Neo4j, neo4jOptions: Neo4jOptions, jobId: String)
       ValidationUtil.isTrue(
         scriptErrors.isEmpty,
         s"""
-           |The following queries inside the `${Neo4jOptions.SCRIPT}` are not valid,
+           |The following queries inside the script options (`${Neo4jOptions.SCRIPT}` / `${Neo4jOptions.SCRIPT_PREFIX}N`) are not valid,
```
ali-ince: same here:


```suggestion
           |The following script queries are not valid,
```

ali-ince (commented):
left some small comments.

src/main/scala/org/neo4j/spark/util/Neo4jOptions.scala
```diff
@@ -365,6 +362,40 @@ class Neo4jOptions(private val options: java.util.Map[String, String]) extends S
...
+          )
+        }
+        (index.toInt, value)
```
nord-erik, line 378: we can still get number format exception here but I think it's fine

nord-erik (approved):

src/main/scala/org/neo4j/spark/util/Neo4jOptions.scala
nord-erik: I remembered it as we decided to keep it to not break existing uses of it / allow more flex for the advanced user.

That said, I would support removing the splitting logic to fix edge cases where someone want's to genuinely use it in their query.

PR #995, opened 2026-06-17
fix: use KEYS as default save strategy

nord-erik:
Our most used save strategy for the connector from what we can tell is
KEYS, not NATIVE. This PR makes the most used scenario the default
scenario.


Emrehzl94 (approved):
💯

PR #1004, opened 2026-06-22
feat: add GDS usage data per transaction

nord-erik:
In order for us to get statistics on how and if GDS are being used from
the spark connector we need at least some clues of what is going on.

This PR will bring in light telemetry for GDS queryTypes

### DIFF
--- a/src/main/scala/org/neo4j/spark/util/Neo4jOptions.scala
+++ b/src/main/scala/org/neo4j/spark/util/Neo4jOptions.scala
@@ -330,7 +330,7 @@ class Neo4jOptions(private val options: Map[String, String]) extends Serializabl
       builder.withTimeout(duration)
     }
 
-    val txMetadata = extractNeo4jTransactionMetadata()
+    val txMetadata = extractConfiguredTransactionMetadata()
     if (query.queryType == QueryType.GDS) {
       txMetadata.put(GDS_TELEMETRY_ACTION, query.value)
     }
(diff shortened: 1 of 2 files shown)


src/main/scala/org/neo4j/spark/util/Neo4jOptions.scala
```diff
@@ -730,6 +735,8 @@ object Neo4jOptions {
 
   // GDS
   private val GDS_OPTION_PREFIX = "gds."
+  val GDS_TELEMETRY_ACTION = "gds.telemetry.action"
+  val GDS_TELEMETRY_NAME = "gds.telemetry.name"
```
fbiville:
```suggestion
  val GDS_TELEMETRY_GRAPH_NAME = "gds.telemetry.graph_name"
```

src/main/scala/org/neo4j/spark/util/Neo4jOptions.scala
```diff
@@ -324,17 +324,21 @@ class Neo4jOptions(private val options: Map[String, String]) extends Serializabl
...
     val timeout = getParameter(TRANSACTION_TIMEOUT_MSECS, DEFAULT_TRANSACTION_TIMEOUT)
-
     val builder = TransactionConfig.builder()
     if (timeout != null) {
       val duration = Duration.ofMillis(timeout.toInt)
       builder.withTimeout(duration)
     }
 
     val txMetadata = extractNeo4jTransactionMetadata()
+    if (query.queryType == QueryType.GDS) {
```
fbiville, line 334: my initial reaction was: "shouldn't this be done inside `extractNeo4jTransactionMetadata`?"
but then I realized that `extractNeo4jTransactionMetadata` is about user-defined tx metadata.
maybe we rename `extractNeo4jTransactionMetadata` into `extractConfiguredTransactionMetadata` or sth?

fbiville (approved):

src/main/scala/org/neo4j/spark/util/Neo4jOptions.scala
nord-erik: yeah I had the same thought process when impl, rename can be valid here

fbiville (approved):

PR #1012, opened 2026-06-24
feat: preserve user-defined query's original return order

fbiville:
The order needs to be preserved as Spark allows users to access DataFrame columns by index.

AutomaticAliaser evolves to not only alias result columns but also extract information of the last RETURN clause.

A new utility called QueryEmbedder leverages that extracted information to create the embedding query's RETURN clause to preserve the order of the original user-defined query.

### DIFF
--- a/src/test/scala/org/neo4j/spark/cypher/QueryEmbedderTest.scala
+++ b/src/test/scala/org/neo4j/spark/cypher/QueryEmbedderTest.scala
@@ -1,0 +1,81 @@
+/*
+ * Copyright (c) "Neo4j"
+ * Neo4j Sweden AB [https://neo4j.com]
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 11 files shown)


src/test/scala/org/neo4j/spark/cypher/QueryEmbedderTest.scala
```diff
@@ -0,0 +1,81 @@
...
+        "uses final return order from nested call",
+        "MATCH (n) CALL (n) { RETURN 1 AS a, 2 AS b } RETURN b AS x, a AS y",
+        "",
+        "CALL () {MATCH (n) CALL (*) {RETURN 1 AS a, 2 AS b} RETURN b AS x, a AS y} RETURN x, y"
+      ),
+      argumentSet(
+        "uses final return names from deeper nested call",
+        "CALL { CALL { RETURN 1 AS a, 2 AS b } RETURN b AS c, a AS d } RETURN d AS x, c AS y",
+        "",
+        "CALL () {CALL () {CALL () {RETURN 1 AS a, 2 AS b} RETURN b AS c, a AS d} RETURN d AS x, c AS y} RETURN x, y"
```
nord-erik, line 72: love it!

nord-erik (approved):
lgtm

ali-ince (approved):

PR #1014, opened 2026-06-26
chore: remove remaining warnings

fbiville:
Manually register Mockito agent to remove the agent auto-registration warning.

Reconfigure assembly plugin to rid of its warnings.

Mute test logs and remove unneeded outputs.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -70,7 +70,8 @@
         <sortpom-maven-plugin.version>4.0.0</sortpom-maven-plugin.version>
         <spark.version>4.1.2</spark.version>
         <spotless-maven-plugin.version>3.7.0</spotless-maven-plugin.version>
-        <test.jvm.args>-XX:+IgnoreUnrecognizedVMOptions
+        <test.jvm.args>
+            -XX:+IgnoreUnrecognizedVMOptions
             --add-opens=java.base/java.lang=ALL-UNNAMED
             --add-opens=java.base/java.lang.invoke=ALL-UNNAMED
             --add-opens=java.base/java.lang.reflect=ALL-UNNAMED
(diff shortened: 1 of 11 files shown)


src/main/scala/org/neo4j/spark/util/Neo4jOptions.scala
```diff
@@ -507,11 +506,9 @@ case class Neo4jDriverOptions(
     }
   }
 
-  @nowarn("cat=deprecation")
   private def toDriverConfig: Config = {
     val builder = Config.builder()
       .withUserAgent(s"neo4j-${Neo4jUtil.connectorEnv}-connector/${Neo4jUtil.connectorVersion}")
-      .withLogging(Logging.slf4j())
```
fbiville, line 518: see comment in `pom.xml`

pom.xml
```diff
@@ -220,6 +229,19 @@
             <version>${connectors-commons.version}</version>
             <scope>runtime</scope>
         </dependency>
+        <dependency>
+            <!-- required for Java driver logging configuration -->
+            <groupId>org.apache.logging.log4j</groupId>
+            <artifactId>log4j-jpl</artifactId>
+            <version>${log4j.version}</version>
+            <scope>test</scope>
```
fbiville, line 249: we could change the scope to runtime, so that's it's similar to the previous (deprecated) driver config: `Logging.slf4j()`

pom.xml
```diff
@@ -646,6 +683,10 @@
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-assembly-plugin</artifactId>
             </plugin>
+            <plugin>
+                <groupId>com.coderplus.maven.plugins</groupId>
+                <artifactId>copy-rename-maven-plugin</artifactId>
```
fbiville, line 694: I initially thought of `maven-antrun-plugin` but that's like killing a fly with a bazooka

fbiville: or maybe we produce 2 JARs, one regular and one shaded, so that solves the card we have about it (ie also produce a regular JAR)

nord-erik (approved):

PR #1023, opened 2026-06-30
feat: add GDS usage data per transaction

nord-erik:
In order for us to get statistics on how and if GDS are being used from
the spark connector we need at least some clues of what is going on.

This PR will bring in light telemetry for GDS queryTypes

### DIFF
--- a/common/src/main/scala/org/neo4j/spark/util/Neo4jOptions.scala
+++ b/common/src/main/scala/org/neo4j/spark/util/Neo4jOptions.scala
@@ -381,6 +381,10 @@ class Neo4jOptions(private val options: java.util.Map[String, String]) extends S
       builder.withTimeout(duration)
     }
 
+    if (query.queryType == QueryType.GDS) {
+      builder.withMetadata(new util.HashMap[String, AnyRef](Map(GDS_TELEMETRY_ACTION -> query.value).asJava))
+    }
+
     builder.build()
   }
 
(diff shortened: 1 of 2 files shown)


common/src/main/scala/org/neo4j/spark/util/Neo4jOptions.scala
```diff
@@ -381,6 +381,10 @@ class Neo4jOptions(private val options: java.util.Map[String, String]) extends S
       builder.withTimeout(duration)
     }
 
+    if (query.queryType == QueryType.GDS) {
+      builder.withMetadata(new util.HashMap[String, AnyRef](Map(GDS_TELEMETRY_ACTION -> query.value).asJava))
```
fbiville, line 385: question: is it possible you override existing metadata here or not?

nord-erik: Actually I think yes. Same applies for the merged 6.0 version?

nord-erik: zooming out a bit, who/what would place existing metadata here, since we're in the builder?

fbiville: ah yep, all good

fbiville (approved):

PR #1028, opened 2026-07-02
test: parallelize test execution

fbiville:
This changes all new IT tests to run concurrently.

Each new IT class now creates a new user, with a custom
home database. That allows the database configuration to
be omitted.

Containers provided by Neo4jContainerProvider start globally
only once.

Tests have been changed to not assume a blank database by default.

This commit introduces a meta-annotation, InjectNeo4jContainerParameter,
which encapsulates various JUnit settings.

### DIFF
--- a/src/main/scala/org/neo4j/spark/streaming/Neo4jStreamingWriter.scala
+++ b/src/main/scala/org/neo4j/spark/streaming/Neo4jStreamingWriter.scala
@@ -17,10 +17,12 @@
 package org.neo4j.spark.streaming
 
 import org.apache.spark.sql.SaveMode
+import org.apache.spark.sql.SparkSession
 import org.apache.spark.sql.connector.write.PhysicalWriteInfo
 import org.apache.spark.sql.connector.write.WriterCommitMessage
 import org.apache.spark.sql.connector.write.streaming.StreamingDataWriterFactory
 import org.apache.spark.sql.connector.write.streaming.StreamingWrite
+import org.apache.spark.sql.streaming.StreamingQueryListener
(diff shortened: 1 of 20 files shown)


src/main/scala/org/neo4j/spark/streaming/Neo4jStreamingWriter.scala
```diff
@@ -65,14 +47,18 @@ class Neo4jStreamingWriter(
...
+      new Neo4jStreamingDataWriterFactory(
+        neo4j,
+        queryId,
+        schema,
+        saveMode,
+        neo4jOptions,
+        scriptResult
+      )
+    } finally {
+      close()
```
fbiville: this seems 100% wrong

nord-erik:
(long comment sorry but for documentation purposes, because will merge with caveat below)

When running locally I see there's a bug with timezones (for tests only). Because writer test cases try to enforce a specific time-zone on the spark session to anchor the time zone reference point.

However for the base test containers we set `TimeZone.getDefault.getID` and I think in a concurrent setting this causes issues. Because the time-zone test cases only fail when I run all tests, suggesting there's leaky state caused by the parallelism.

Affected test is ReadIT#returns_selected_time_column

```
Expected :"12:23:00.294Z"
Actual   :"12:23:00.294+02:00"
```

My best guesstimate is that this will either go away or be easier to fix once we refactor WriteIT. And refactor to WriteIT will be easier with this in.

Suggested fix would be to anchor the test container to a static-determined time zone, and anchor the session to a static-determined time-zone. Then adjust assertions based on that.

nord-erik (approved):

PR #1031, opened 2026-07-03
docs: fix playbook sources

recrwplay:
The docs playbook files need updating to allow for the fact that the docs are in a subdirectory of the repo.

### DIFF
--- a/docs/publish.yml
+++ b/docs/publish.yml
@@ -7,14 +7,8 @@ site:
 content:
   sources:
   - url: ../
-    edit_url: https://github.com/neo4j/docs-spark/tree/{refname}/{path}
-    branches: HEAD
+    branches: ['HEAD', '4.2', '4.1', '4.0']
     start_path: docs
-    include: docs
-      exclude:
-      - '!**/_includes/*'
-      - '!**/readme.adoc'
-      - '!**/README.adoc'
 
 ui:
   bundle:
--- a/docs/publish.yml
+++ b/docs/publish.yml
@@ -6,9 +6,11 @@ site:
 
 content:
   sources:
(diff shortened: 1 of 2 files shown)


docs/publish.yml
```diff
@@ -6,9 +6,11 @@ site:
 
 content:
   sources:
-    - url: ./
-      edit_url: https://github.com/neo4j/docs-spark/tree/{refname}/{path}
-      branches: ['HEAD', '4.2', '4.1', '4.0']
+  - url: ../
+    edit_url: https://github.com/neo4j/docs-spark/tree/{refname}/{path}
```
nvitucci: Should we update this to the new repo?

docs/publish.yml
```diff
@@ -6,9 +6,11 @@ site:
...
+    branches: HEAD
+    start_path: docs
+    include: docs
       exclude:
       - '!**/_includes/*'
       - '!**/readme.adoc'
```
nvitucci: I believe the keys here are case-insensitive so it _should_ catch our `README.adoc`, but I am not 100% sure.

docs/publish.yml
recrwplay: We can actually just remove it because the 'raise an issue' link, when it is displayed, should be automatically calculated.

docs/publish.yml
recrwplay: I don't think they even serve a purpose - Antora doesn't read include or exclude here as far as I can tell. I've deleted them.

nord-erik:
Let us know if there's anything here that needs our attention!

nvitucci (approved):

PR #1038, opened 2026-07-14
refactor: introduce WriteIT, replacing DataSourceWriterTSE

nord-erik:
Collects `DataSourceWriterTSE`  test cases into a test suite that is
then executed concurrently with the other *IT suites.

### concurrency structure quirk
Every nested test class share the same database, both instance
and logical database. Therefore it is not possible to reset the
database in a concurrency safe manner because resetting a
database will for sure affect another test case...

For nested cases where only `@Test` is used it is possible to
run the whole nested test class sequentially (SAME_THREAD)
and allow for resets. But it is not possible to have dynamic
tests to run sequentially.

Because dynamic tests cannot reset the database safely, this
means each dynamic test need to do unique (within test class)
writes either via unique labels or unique relationship types.

### SparkSession.setActiveSessions spam
The connector picks up the default SparkSession from the current
thread by default. But because we run concurrently, there is a pretty
big chance that thread A creates a DF in e.g. a test factory set-up,
then thread B performs the connector operation from a dynamic
test case runtime.

Because of this explicit spark sessions must be set inside each unit
of work (dynamic test or test).


nord-erik:
Sorry did not mean to open as a PR, meant to open as a draft

nord-erik:
Current state:

|Spark/SQL/JVM	| Neo4j		| Tested	| Array		|
|---------------|---------------|---------------|---------------|
| string	| String	| Scala, SQL	| Scala, SQL	|
| long		| Integer	| Scala, SQL	| Scala, SQL	|
| integer	| Integer	| Scala, SQL	| Scala, SQL	|
| short		| Integer	| Scala, SQL	| Scala, SQL	|
| byte		| Integer	| Scala, SQL	| Scala, SQL	|
| real		| Float		| SQL		| SQL		|
| float		| Float		| Scala, SQL	| Scala, SQL	|
| double	| Float		| Scala, SQL	| Scala, SQL	|
| decimal	| String	| Scala, SQL	| Scala, SQL	|
| boolean	| Boolean	| Scala, SQL	| Scala, SQL	|
| date		| Date		| Scala, SQL	| Scala, SQL	|
| timestamp	| ZonedDateTime	| Scala, SQL	| Scala, SQL	|
| timestamp_ntz	| LocalDateTime	| Scala, SQL	| Scala, SQL	|
| duration	| Duration	| Scala		| Scala		|
| period	| Duration	| Scala		| Scala		|
| INTERVAL	| Duration	| SQL		| SQL		|
| array[byte]	| ByteArray	| Scala, SQL	| Scala, SQL	|
| struct	| Point		| Scala		| Scala		|
| struct	| LocalTime	| Scala		| Scala		|
| struct	| OffsetTime	| Scala		| Scala		|
| struct	| Duration	| Scala		| Scala		|
| Point		| Point		| undocumented	| undocumented	|
| LocalTime	| LocalTime	| undocumented	| undocumented	|
| OffsetTime	| OffsetTime	| undocumented	| undocumented	|

nord-erik:
I found that in the original test cases we support Point, LocalTime and OffsetTime objects natively even though the documentation say we only support them via structs

nord-erik:
Related bug (I encounter this in my env)

https://youtrack.jetbrains.com/projects/IDEA/issues/IDEA-389426/Randomly-ignored-Tests-on-Parallel-Test-Execution

nord-erik:
Removed cases:

| Test case | Reason |
|-----------|--------|
| should write nodes with map values into Neo4j | this case tests driver/scala integraion not spark/neo4j integration |
| shouldWriteComplexDF | this test case is just raw cypher in a convoluted way |
| should work create source node and match target node | under assumption that test case should_overwrite_related_graph_entities_when_node_overwrite_mode pass and should_handle_relationship_match_source_even_when_odd_characters_present pass and under the assumption that we test Neo4jQueryService, this test case is now testing Cypher. because we've already confirmed that Neo4jQueryService is reached via the other test cases |
| should work match source node and merge target node | is just a specific case of should work match source node and merge target node with odd chars |
| should work match source node and merge target node with odd chars | replaced by should_handle_relationship_match_source_even_when_odd_characters_present |

src/test/scala/org/neo4j/spark/WriteIT.scala
```diff
@@ -0,0 +1,1544 @@
...
+        ("hammer", "screw", false),
+        ("driver", "nail", false)
+      ).toDF("tööl", "part:specification", "is_appropriate")
+        .repartition()
+        .write.format("neo4j")
+        .mode(SaveMode.Overwrite)
+        .option("relationship", "INTERACTS_WITH")
+        .option("relationship.source.save.mode", "match")
+        .option("relationship.source.labels", ":Tool")
+        .option("relationship.source.node.keys", "tööl:name")
```
Emrehzl94: ö is not odd to me :D

Emrehzl94 (approved):
lgtm!

src/test/scala/org/neo4j/spark/WriteIT.scala
nord-erik: Same hehe can use ‽ and õ

vasifabdulhamidov-neo (approved):

nord-erik:
This PR struggles with flakyness in CI but stable on my machine......

When time allows I will have a look!

PR #1042, opened 2026-07-16
ci(docs): extend doc artifact validation rules

venikkin:
Potential fix for [https://github.com/neo4j/neo4j-spark-connector/security/code-scanning/1](https://github.com/neo4j/neo4j-spark-connector/security/code-scanning/1)

General fix: treat workflow-run artifacts as untrusted, extract only into isolated temp dirs, and strictly validate archive entries before extraction/use (reject absolute paths, traversal, symlinks, and unexpected top-level content). Then deploy only validated content.

Best targeted fix in this file: harden the existing `suspicious-path-check` step so it performs robust zip entry validation (not just `../`). Specifically, in `.github/workflows/docs-deploy-surge.yml`, replace the current `unzip -l ... | grep "\.\./"` check with a loop over `zipinfo -1 docs.zip` entries that rejects:
- absolute paths (`^/`)
- traversal segments (`(^|/)\.\.(/|$)`)
- backslashes (`\`)  
Also fail if no entries exist. This keeps behavior (deploy docs artifact) while preventing poisoned archive paths from being extracted/propagated.

### DIFF
--- a/.github/workflows/docs-deploy-surge.yml
+++ b/.github/workflows/docs-deploy-surge.yml
@@ -86,13 +86,32 @@ jobs:
 
       - id: suspicious-path-check
         name: Suspicious paths check
+        shell: bash
         env:
           ARTIFACT_DIR: ${{ runner.temp }}/artifacts/docs
         run: |
+          set -euo pipefail
           cd "$ARTIFACT_DIR"
-          if unzip -l docs.zip | grep -q "\.\./"; then
+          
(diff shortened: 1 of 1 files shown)


neo4j-docops-agent:
Thanks for the documentation updates.

The preview documentation has now been torn down - reopening this PR will republish it.

nord-erik (approved):
nice!

PR #1044, opened 2026-07-16
feat: publish a non-shaded spark-slim artifact

Emrehzl94:
Splits the connector into submodules so we can ship two artifacts:

- org.neo4j.connectors:spark — the shaded/fat jar (unchanged coordinate and behavior)
- org.neo4j.connectors:spark-slim — new non-shaded jar

**Structure**
- spark-parent (pom) holds shared properties, dependency and plugin management
- spark-slim(jar) owns the source, dependencies, and tests - produces the thin jar with its own pom declaring dependencies
- spark(jar) depends on spark-slim and assembles the fat jar

**Important**
- JReleaser publishes both to Maven Central and attaches the fat + slim jars as GitHub release assets
- spark reuses spark-slim's sources/javadoc (via build-helper) to satisfy Maven Central's artifact requirements
- Fixed deploy (assembly attach=false)
- README updated with the new coordinates


spark-slim/src/main/scala/org/neo4j/spark/converter/TypeConverter.scala
```diff
@@ -16,11 +16,7 @@
  */
 package org.neo4j.spark.converter
 
-import org.apache.spark.sql.types.DataType
-import org.apache.spark.sql.types.DataTypes
-import org.apache.spark.sql.types.DayTimeIntervalType
-import org.apache.spark.sql.types.DecimalType
-import org.apache.spark.sql.types.YearMonthIntervalType
+import org.apache.spark.sql.types._
```
venikkin: I don't think it's a blocker, but usually it's good practice to explicitly list imports to avoid possible clashes on dependency updates. I don't think spotless enforces this, so most likely it comes from the IDE, so maybe we need to add .editconfig to this repo (although it's out of the scope now).

venikkin (approved):

spark-slim/src/main/scala/org/neo4j/spark/converter/TypeConverter.scala
Emrehzl94: yeah thanks good catch 👍

PR #1053, opened 2026-07-22
test: disable parallel tests

venikkin:
To reduce load on build agents.


nord-erik (approved):
lets do this temporarity, longer running times but with increased stability

we're so close there's basically StreamingIT + TransactionTimeoutIT (potentially fixed) cases that are flaky for various reasons...

PR #1063, opened 2026-07-27
ci: docs changes do not trigger TC

nord-erik:
This PR introduce a new trigger filter aimed to skip team city builds
if and only if the new incoming change only touches documentation files


nord-erik:
Cherry picked from 6.0: https://github.com/neo4j/neo4j-spark-connector/pull/1062

dhrudevalia (approved):

PR #1068, opened 2026-07-30
ci: short-circuit pipeline when changes are docs only

ali-ince:
This pull request introduces a mechanism to skip unnecessary CI build steps for documentation-only changes, optimizing build times and resource usage. It does this by detecting when a change only affects documentation files and conditionally skipping most build steps while still reporting commit status to GitHub. Additionally, it removes redundant documentation-specific trigger rules from the TeamCity configuration.

This is because branch protection rules still expect `complete` build to be reported as successful, regardless of whether the PR is a docs only change or not.


nord-erik (approved):
Let's have a go with this approach

PR #1080, opened 2026-08-04
fix: race condition in DriverCache

nord-erik:
Previous implementation of DriverCache suffered from a check-then-act
race condition, where retrieving drivers from cache and updating the
counter was two separate atomic actions. This can cause race condition
under this scenario

Imagine thread A `close()`. Right after it calls `cache.get()` followed
by `decrementAndGet() == 0`. If the thread gets context-switched before
it get's the chance to `cache.remove()`, the race error happen.

What will happen in thread B, it can `getOrCreate()`, which will then
get the same driver as thread A is about to remove and close. It will
also increment the counter.

Back to A, it will now remove the live driver from cache and close the
driver! This will cause Thread B's driver to be stale. It will also
cause future calls to create a duplicated driver (cache empty).

This fix aims to make counters and accesses into the same atomic action
for the DriverCache. Preventing the above scenario.

Cherry-pick from 88077eedc3095a8be1cb6aa992240829b825ea22

### DIFF
--- a/common/src/main/scala/org/neo4j/spark/util/DriverCache.scala
+++ b/common/src/main/scala/org/neo4j/spark/util/DriverCache.scala
@@ -17,32 +17,47 @@
 package org.neo4j.spark.util
 
 import org.neo4j.driver.Driver
-import org.neo4j.spark.util.DriverCache.cache
 
 import java.util.concurrent.ConcurrentHashMap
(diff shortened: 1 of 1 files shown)


vasifabdulhamidov-neo (approved):
Lgtm

PR #1083, opened 2026-08-05
test: spark declarative pipeline integration test

nord-erik:
This commit introduces a new PySpark integration test where we use the
new Spark 4-introduced declarative pipeline in conjunction with the
Neo4j spark connector.

Keeping it simple. The actual `test_spark_declarative_pipeline.py`
script only orchestrates the test harness. Then, the pipeline itself
is a simple case of using the connector to find relationships in
Neo4j via Spark Declarative Pipeline specification.

The test harness creates the required directory structure inside a
temporary directory. Using templates so we can configure the
pipeline to target the test container.

Then it runs spark-pipeline and checks for any errors. Assertions
are done as part of the pipeline.


scripts/python/test_spark.py
nord-erik, line 1: formatting imports only, my editor does it automatically,

scripts/python/pipeline/test_spark_declarative_pipeline.py
```diff
@@ -0,0 +1,288 @@
...
+        major_version = int(version_parts[0])
+        minor_version = int(version_parts[1])
+    except (IndexError, ValueError) as error:
+        raise IntegrationTestError(f"PySpark version parse error: {version}") from error
+
+    if not major_version >= expected_major or not minor_version >= expected_minor:
+        raise IntegrationTestError(f"Require at lest version 4.1.x, found: {version}")
+
+
+def find_spark_pipelines_executable() -> Path:
```
nord-erik: With full knowledge of CI environment this can be deleted and hard coded

But as for now I don't know where the spark pipeline executable will be in CI

scripts/python/pipeline/neo4j_pipeline_template.py
```diff
@@ -0,0 +1,96 @@
+from pyspark import pipelines as dp
+from pyspark.sql import DataFrame, SparkSession
+from pyspark.sql import functions as F
+
+spark = SparkSession.active()
+
+NEO4J_URL = "%%%__NEO4J_URL__%%%"
```
ali-ince: could this be looked up through environment instead of templating?

scripts/python/pipeline/spark-pipeline-template.yml
```diff
@@ -0,0 +1,7 @@
+name: neo4j_spark_declarative_relationship_integration_test
```
ali-ince: I see in the [docs](https://spark.apache.org/docs/latest/declarative-pipelines-programming-guide.html#spark-pipelines-init) that there's this `spark-pipelines init --name xyz` command that builds this file, excluding the `spark.sql.shuffle.partitions` configuration. If that's not that important for us, maybe it will be cleaner to utilize that command and just get rid of the templating logic completely (also see my previous comment on `neo4j_pipeline_template.py` file?

scripts/python/pipeline/test_spark_declarative_pipeline.py
ali-ince: I think we could get rid of this, as `pip` should deal with resolving this properly. If we want to be explicit, then we could also consider using `venv` to create a virtual python environment per test.

ali-ince (changes requested):
Looks good in general. I left a few suggestions which I believe will simplify the test structure.

scripts/python/pipeline/test_spark_declarative_pipeline.py
nord-erik: I'm doing venv locally, didn't occur to me to do it in CI as well hehe, but we should for consistency.

scripts/python/pipeline/neo4j_pipeline_template.py
nord-erik: I'll look into it, the script is submitted to a spark binary. Not sure _what_ environment can be read once pipeline runtime kicks in.

scripts/python/pipeline/spark-pipeline-template.yml
nord-erik: I'll have a look

scripts/python/pipeline/test_spark_declarative_pipeline.py
nord-erik: Skipping venv right now because we use mise and pip, should be equivalent enough. Removed the probing function and instead simply calling the pipelines executable

ali-ince (approved):
💯

PR #1099, opened 2026-08-25
docs: cover when streaming can cause data loss

nord-erik:
In a recent audit we found that actually, if you put something trivial
as your streaming property for Spark structured stream reading, you
might end up with data loss.

Because if you write non-unique but time ordered property values, you
might miss out on data being written to Neo4j while you are mid-stream
reading.

This PR aims to improve the docs to warn about this race condition and
makes the requirements for a good property clearer to understand.

### DIFF
--- a/docs/modules/ROOT/pages/streaming.adoc
+++ b/docs/modules/ROOT/pages/streaming.adoc
@@ -152,8 +152,8 @@ The `streaming.property.name` option sets the name of the node or relationship p
 [WARNING]
 ====
 If the values of your streaming property are not unique, you might lose data.
-Structured streaming reads keep track of the progress using the value of the streaming property, which must increase for each subsequent read.
-If this value is already present in a previously written record, the new data will not be read.
+Structured streaming reads keep track of the progress using the value of the streaming property, which must increase for each subsequent read to keep order.
+If you write a new record during streaming whose streaming property value is less than the current stream's checkpoint, that record will not bea read.
 ====
 
 Although the Neo4j `timestamp()` function might seem a good choice to create a streaming property, its millisecond resolution may not be enough to guarantee uniqueness in a write-heavy scenario.
--- a/docs/modules/ROOT/pages/streaming.adoc
+++ b/docs/modules/ROOT/pages/streaming.adoc
@@ -143,20 +143,26 @@ spark \
 
 === Streaming property name
 
-For the streaming to work, you need each record to have a property of type `timestamp`
-to leverage when reading new data from Neo4j to be sent to the stream.
-
-Behind the scenes the connector is building a query with a `WHERE` clause that checks for the
-records that have this `[timestampProperty]` between a range of timestamps computed from checkpoint data
-and latest offset available in the database.
-
-So it's required that each node has the timestamp property of a Neo4j type (Long),
-and it *must* be not `null`.
-
-[NOTE]
-A property of type string like "2021-08-11" does not work. It needs to be a Long of Neo4j type.
-
+For the streaming reads, you need each record to have a "streaming property" to leverage when reading new data from Neo4j to be sent to the stream.
+The name of your streaming property must be passed along using the required option `streaming.property.name`.
 The property name can be anything, just remember to set the `streaming.property.name` accordingly.
+
+Behind the scenes the connector is building a query with a `WHERE` clause that checks for the records' streaming property that falls between a range of values computed from checkpoint data and latest offset available in the database.
+
+Because of how we construct this `WHERE` clause, it is _required_ to use the `Integer` type for the property (equivalent to Java/Scala's Long).
+It must not be `null`.
+We recommend that you use time-ordered sequence of numbers that are uniquely identifying (e.g. an unique incrementing ID).
+
+[WARNING]
+====
+If the values of your streaming property are not unique, you might lose data.
+Structured streaming reads keep track of progress using the highest property seen, then move on to a value strictly greater than that next time.
+If you write a record (mid-streaming) that has a streaming property which is a duplicate of an already read record, this data will not be streamed/read.
+====
+
+Using Neo4j built-in function `timestamp()` as value for these properties is a natural instinct.
+But be warned, if you have records with non-unique timestamps you might have duplicated timestamps in your data.
+Of course you can still use e.g. a `timestamp()` value if your domain can guarantee that this is unique or the above warning is not of concern.
 
 === _Streaming from_ option
 
--- a/docs/modules/ROOT/pages/streaming.adoc
+++ b/docs/modules/ROOT/pages/streaming.adoc
@@ -162,7 +162,10 @@ If you write a record (mid-streaming) that has a streaming property which is a d
 
 Using Neo4j built-in function `timestamp()` as value for these properties is a natural instinct.
 But be warned, if you have records with non-unique timestamps you might have duplicated timestamps in your data.
-Of course you can still use e.g. a `timestamp()` value if your domain can guarantee that this is unique or the above warning is not of concern.
+Imagine you are reading a stream from a Neo4j database under a heavy write scenario over multiple transactions.
+Every timestamp value of writes that happens within the same millisecond will have non-unique timestamp, and is prone to the above warning.
+A streaming read might reach Neo4j in-between two write transactions within the same millisecond.
+The later written records will then never be read.
 
 === _Streaming from_ option
 
--- a/docs/modules/ROOT/pages/streaming.adoc
+++ b/docs/modules/ROOT/pages/streaming.adoc
@@ -153,7 +153,7 @@ The `streaming.property.name` option sets the name of the node or relationship p
 ====
 If the values of your streaming property are not unique, you might lose data.
 Structured streaming reads keep track of the progress using the value of the streaming property, which must increase for each subsequent read to keep order.
-If you write a new record during streaming whose streaming property value is less than the current stream's checkpoint, that record will not bea read.
+If you write a new record during streaming whose streaming property value is less than the current stream's checkpoint, that record will not be read.
 ====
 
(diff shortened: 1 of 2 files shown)


neo4j-docops-agent:
This PR includes documentation updates
View the updated docs at https://neo4j-neo4j-spark-connector-1099.surge.sh

Updated pages:
- [Spark Structured Streaming](https://neo4j-neo4j-spark-connector-1099.surge.sh/spark/5/streaming)

docs/modules/ROOT/pages/streaming.adoc
```diff
@@ -143,20 +143,26 @@ spark \
...
-and it *must* be not `null`.
+Because of how we construct this `WHERE` clause, it is _required_ to use the `Integer` type for the property (equivalent to Java/Scala's Long).
+It must not be `null`.
+We recommend that you use time-ordered sequence of numbers that are uniquely identifying (e.g. an unique incrementing ID).
 
-[NOTE]
-A property of type string like "2021-08-11" does not work. It needs to be a Long of Neo4j type.
+[WARNING]
+====
+If the values of your streaming property is not uniquely identifying, you might lose data.
```
fbiville:
```suggestion
If the values of your streaming property are not unique, you might lose data.
```

docs/modules/ROOT/pages/streaming.adoc
```diff
@@ -143,20 +143,26 @@ spark \
...
+Structured streaming reads keeps track of progress using the highest property seen, then reads strictly greater than it next time.
```
fbiville:
```suggestion
Structured streaming reads keep track of progress using the highest property seen, then move on to a value strictly greater than that next time.
```

docs/modules/ROOT/pages/streaming.adoc
```diff
@@ -143,20 +143,26 @@ spark \
...
+If you write a record (mid-streaming) that has a streaming property which is a duplicate of an already read record, this data will not be streamed/read.
+====
 
-The property name can be anything, just remember to set the `streaming.property.name` accordingly.
+Using Neo4j built-in function `timestamp()` as value for these properties is a natural instinct.
+But be warned, if you have records with non-unique timestamps you might have duplicated timestamps in your data.
```
fbiville: suggestion: I'd be more explicit about the timestamp precision and how it can end up with duplicates (+ `timestamp()` returns always the same value within 1 transaction iirc)

nord-erik: thanks, didn't know about that transaction caveat, if that's the case I can easily see how one would end up with many duplicated timestamps

nord-erik: reading some code this morning I think if all the duplicated stamps come in as one transaction, this "race" condition doesn't happen, the streaming read will fetch all of them.

If two transactions with duplicated/overlapping streaming property values would commit at two different times, that's when we have a problem. Didn't want this docs to be overly in-depth technical so skipped articulating those details.

neo4j-docops-agent:
This PR includes documentation updates
View the updated docs at https://neo4j-neo4j-spark-connector-1099.surge.sh

Updated pages:
- [Spark Structured Streaming](https://neo4j-neo4j-spark-connector-1099.surge.sh/spark/5/streaming)

docs/modules/ROOT/pages/streaming.adoc
fbiville: I think we can keep it simple enough by mentioning "heavy write scenarios" with multiple transactions happening within the same millisecond?

nord-erik: sounds good

fbiville (approved):

docs/modules/ROOT/pages/streaming.adoc
```diff
@@ -143,20 +143,29 @@ spark \
...
+It must not be `null`.
+We recommend that you use time-ordered sequence of numbers that are uniquely identifying (e.g. an unique incrementing ID).
 
-[NOTE]
-A property of type string like "2021-08-11" does not work. It needs to be a Long of Neo4j type.
+[WARNING]
+====
+If the values of your streaming property are not unique, you might lose data.
+Structured streaming reads keep track of progress using the highest property seen, then move on to a value strictly greater than that next time.
+If you write a record (mid-streaming) that has a streaming property which is a duplicate of an already read record, this data will not be streamed/read.
```
nvitucci:
```suggestion
Structured streaming reads keep track of the progress using the value of the streaming property, which must increase for each subsequent read.
If this value is already present in a previously written record, the new data will not be read.
```

docs/modules/ROOT/pages/streaming.adoc
```diff
@@ -143,20 +143,29 @@ spark \
 
 === Streaming property name
```
nvitucci:
```suggestion
=== Streaming property
```

docs/modules/ROOT/pages/streaming.adoc
```diff
@@ -143,20 +143,29 @@ spark \
...
-Behind the scenes the connector is building a query with a `WHERE` clause that checks for the
-records that have this `[timestampProperty]` between a range of timestamps computed from checkpoint data
-and latest offset available in the database.
+Behind the scenes the connector is building a query with a `WHERE` clause that checks for the records' streaming property that falls between a range of values computed from checkpoint data and latest offset available in the database.
 
-So it's required that each node has the timestamp property of a Neo4j type (Long),
-and it *must* be not `null`.
+Because of how we construct this `WHERE` clause, it is _required_ to use the `Integer` type for the property (equivalent to Java/Scala's Long).
+It must not be `null`.
+We recommend that you use time-ordered sequence of numbers that are uniquely identifying (e.g. an unique incrementing ID).
```
nvitucci:
```suggestion
When reading from Neo4j, each record must include a unique increasing `Integer` value (a _streaming property_).

The connector uses the value of the streaming property to filter records between the checkpoint and the latest offset stored in the database.

The `streaming.property.name` option sets the name of the node or relationship property to use.
```

docs/modules/ROOT/pages/streaming.adoc
```diff
@@ -143,20 +143,29 @@ spark \
...
+If you write a record (mid-streaming) that has a streaming property which is a duplicate of an already read record, this data will not be streamed/read.
+====
 
-The property name can be anything, just remember to set the `streaming.property.name` accordingly.
+Using Neo4j built-in function `timestamp()` as value for these properties is a natural instinct.
+But be warned, if you have records with non-unique timestamps you might have duplicated timestamps in your data.
+Imagine you are reading a stream from a Neo4j database under a heavy write scenario over multiple transactions.
+Every timestamp value of writes that happens within the same millisecond will have non-unique timestamp, and is prone to the above warning.
+A streaming read might reach Neo4j in-between two write transactions within the same millisecond.
+The later written records will then never be read.
```
nvitucci:
```suggestion
Although the Neo4j `timestamp()` function might seem a good choice to create a streaming property, its millisecond resolution may not be enough to guarantee uniqueness in a write-heavy scenario.
If several transactions happen within the same millisecond, the connector will only read one.
```

nvitucci (changes requested):
I've rephrased it a bit to simplify and make it more fluent. Let me know if it still makes sense for the issue you want to highlight.

docs/modules/ROOT/pages/streaming.adoc
nord-erik: I still think we need to include the guideline of how to pick a good value

We need this bit: "We recommend that you use time-ordered sequence of numbers that are uniquely identifying (e.g. an unique incrementing ID)."

But can try to fit it somewhere appropriate in the newer version,

docs/modules/ROOT/pages/streaming.adoc
nord-erik: I'm taking it in but will add a change for next round of review

The values that are not read will be records that have a `streaming.property.name` that is "older" that the latest checkpoint mid-stream.

So say you stream A -> B -> D -> E -> F

If you insert "C" when the reader is at D, that's the issue. It will not back-track and read C, it already skipped it.

docs/modules/ROOT/pages/streaming.adoc
nord-erik: Disregard my comment its right there in the beginning of the diff :D

neo4j-docops-agent:
This PR includes documentation updates
View the updated docs at https://neo4j-neo4j-spark-connector-1099.surge.sh

Updated pages:
- [Spark Structured Streaming](https://neo4j-neo4j-spark-connector-1099.surge.sh/spark/5/streaming)

neo4j-docops-agent:
This PR includes documentation updates
View the updated docs at https://neo4j-neo4j-spark-connector-1099.surge.sh

Updated pages:
- [Spark Structured Streaming](https://neo4j-neo4j-spark-connector-1099.surge.sh/spark/5/streaming)

docs/modules/ROOT/pages/streaming.adoc
```diff
@@ -141,22 +141,23 @@ spark \
...
-and it *must* be not `null`.
+The `streaming.property.name` option sets the name of the node or relationship property to use.
 
-[NOTE]
-A property of type string like "2021-08-11" does not work. It needs to be a Long of Neo4j type.
+[WARNING]
+====
+If the values of your streaming property are not unique, you might lose data.
+Structured streaming reads keep track of the progress using the value of the streaming property, which must increase for each subsequent read to keep order.
+If you write a new record during streaming whose streaming property value is less than the current stream's checkpoint, that record will not bea read.
```
nvitucci:
```suggestion
If you write a new record during streaming whose streaming property value is less than the current stream's checkpoint, that record will not be read.
```

nvitucci (approved):

PR #1101, opened 2026-08-26
fix: handle alwaysTrue and alwaysFalse

nord-erik:
Neo4jImplicits are capable of creating `AwlaysTrue` and `AlwaysFalse`
filters. However Neo4jUtil considered this as unsupported filter types.

We should be able to safely map `AlwaysTrue` literal true and
`AlwaysFalse` to literal false.

### DIFF
--- a/common/src/main/scala/org/neo4j/spark/util/Neo4jUtil.scala
+++ b/common/src/main/scala/org/neo4j/spark/util/Neo4jUtil.scala
@@ -198,10 +198,13 @@ object Neo4jUtil {
       case contains: StringContains =>
         getCorrectProperty(container, attributeAlias.getOrElse(contains.attribute))
           .contains(valueToCypherExpression(contains.attribute, contains.value))
-      case notNull: IsNotNull   => getCorrectProperty(container, attributeAlias.getOrElse(notNull.attribute)).isNotNull
-      case isNull: IsNull       => getCorrectProperty(container, attributeAlias.getOrElse(isNull.attribute)).isNull
-      case not: Not             => mapSparkFiltersToCypher(not.child, container, attributeAlias).not()
-      case filter @ (_: Filter) => throw new IllegalArgumentException(s"Filter of type `$filter` is not supported.")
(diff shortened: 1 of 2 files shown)


common/src/main/scala/org/neo4j/spark/util/Neo4jUtil.scala
```diff
@@ -198,10 +198,13 @@ object Neo4jUtil {
...
         getCorrectProperty(container, attributeAlias.getOrElse(contains.attribute))
           .contains(valueToCypherExpression(contains.attribute, contains.value))
-      case notNull: IsNotNull   => getCorrectProperty(container, attributeAlias.getOrElse(notNull.attribute)).isNotNull
-      case isNull: IsNull       => getCorrectProperty(container, attributeAlias.getOrElse(isNull.attribute)).isNull
-      case not: Not             => mapSparkFiltersToCypher(not.child, container, attributeAlias).not()
-      case filter @ (_: Filter) => throw new IllegalArgumentException(s"Filter of type `$filter` is not supported.")
+      case notNull: IsNotNull => getCorrectProperty(container, attributeAlias.getOrElse(notNull.attribute)).isNotNull
+      case isNull: IsNull     => getCorrectProperty(container, attributeAlias.getOrElse(isNull.attribute)).isNull
+      case not: Not           => mapSparkFiltersToCypher(not.child, container, attributeAlias).not()
+      case _: AlwaysTrue      => Conditions.noCondition()
```
Emrehzl94: why didn't we do this `Cypher.literalTrue().asCondition()`?

nord-erik: Honestly because of AI suggestion.

But I think this case represents "if (true == true)" so simply adding no condition is equivalent

Next case is "if (true == false)" so will always boil down to literal false

nord-erik: Let's change it to `Cypher.literalTrue().asCondition()` so that negation works, thanks for spotting

Emrehzl94 (approved):
💯

PR #1105, opened 2026-08-27

nord-erik:

identifier names.

Just using Cypher DSL v2022 creates regressions around two areas:
 - Sanitize values that start with "$" e.g. "$string"

Implemented in such way where these quirks are retained to keep
backwards compatibility.

### DIFF
--- a/common/src/main/scala/org/neo4j/spark/util/Neo4jImplicits.scala
+++ b/common/src/main/scala/org/neo4j/spark/util/Neo4jImplicits.scala
@@ -46,6 +46,7 @@ import org.apache.spark.sql.types.DataTypes
 import org.apache.spark.sql.types.MapType
 import org.apache.spark.sql.types.StructField
 import org.apache.spark.sql.types.StructType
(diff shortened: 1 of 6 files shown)


common/src/main/scala/org/neo4j/spark/util/Neo4jImplicits.scala
```diff
@@ -56,18 +57,27 @@ import org.neo4j.spark.service.SchemaService
...
-import javax.lang.model.SourceVersion
-
 object Neo4jImplicits {
 
+  private val BACKTICK = "`"
+
   implicit class CypherImplicits(str: String) {
-    private def isValidCypherIdentifier() = SourceVersion.isIdentifier(str) && !str.trim.startsWith("$")
 
-    def quote(): String = if (!isValidCypherIdentifier() && !str.isQuoted()) s"`$str`" else str
```
nord-erik, line 66: function no longer quotes, it sanitizes, hence rename

vasifabdulhamidov-neo (approved):
👍

PR #1112, opened 2026-08-31
fix: labels containing spaces causes slowdown

nord-erik:
Before, we sanitized/escaped/quoted the key used to look-up the
returned map from `apoc.meta.stats()`. However, Cypher produces these
keys as non-escaped string values. So if we escape the predicted key,
we will never match the produced APOC key.

The symptom of this state is: instead of error and fall back to count
query, we use Long.Max value for each executor's pagination. Causing
each worker thread to scan the whole graph, causing slowdown.

This PR fixes the error state such that we do not escape the key.
In the case where no matches was found, we fall back to query.

### DIFF
--- a/common/src/main/scala/org/neo4j/spark/service/SchemaService.scala
+++ b/common/src/main/scala/org/neo4j/spark/service/SchemaService.scala
@@ -518,21 +518,23 @@ class SchemaService(
           .asInstanceOf[util.Map[String, Long]]
           .asScala
 
-        val minFromSource = options.relationshipMetadata.source.labels // no sanitization here - labels used as map keys
+        val sourceCounts = options.relationshipMetadata.source.labels // no sanitization here - labels used as map keys
           .map(label =>
             map.get(s"(:$label)-[:${options.relationshipMetadata.relationshipType}]->()")
           )
 
-        val minFromTarget = options.relationshipMetadata.target.labels // no sanitization here - labels used as map keys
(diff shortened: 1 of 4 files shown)


venikkin (commented):
Looks good. I think we need to add a test covering at least that we can find count relationship with nodes with labels that mighe be corrupted by sanitation.

common/src/main/scala/org/neo4j/spark/service/SchemaService.scala
```diff
@@ -517,19 +517,23 @@ class SchemaService(
...
+
+        val minFromTarget = options.relationshipMetadata.target.labels // no sanitization here - labels used as map keys
           .map(label =>
-            map.get(s"()-[:${options.relationshipMetadata.relationshipType}]->(:$label)").getOrElse(Long.MaxValue)
+            map.get(s"()-[:${options.relationshipMetadata.relationshipType}]->(:$label)")
           )
-          .min
-        Math.min(minFromSource, minFromTarget)
+
+        if (minFromSource.contains(None) || minFromTarget.contains(None)) {
```
vasifabdulhamidov-neo: should we also check list emptiness?

nord-erik: Good catch Vasif! Newest version changes the logic so that we capture the empty case as well, please see

venikkin (approved):

PR #1117, opened 2026-09-01
fix: resolve an ambiguity between alias separation and dots in property names

venikkin:
This bug is related to all properties containing dot, not only packaged map structs. 

The existing implementation treats dot primarily as separator between an alias and a property, and strips it when reading. For example flattened map property `mymap.myproperty` becomes `myproperty` and thus always read as null. 

In reality, the dot has 3 possible meanings: 
* an alias
* part of the property 
* nested field access (for example in `points`). 

This PR addresses this ambiguity. It makes alias stripping explicit and requires the exact alias name, where possible. Filter encoding also uses extra backticks to distinguish between property name and field access.

### DIFF
--- a/src/main/scala/org/neo4j/spark/service/Neo4jQueryService.scala
+++ b/src/main/scala/org/neo4j/spark/service/Neo4jQueryService.scala
@@ -26,14 +26,7 @@ import org.apache.spark.sql.sources.And
 import org.apache.spark.sql.sources.Filter
 import org.apache.spark.sql.sources.Or
 import org.neo4j.caniuse.Neo4j
-import org.neo4j.cypherdsl.core.Cypher.callRawCypher
 import org.neo4j.cypherdsl.core._
-import org.neo4j.cypherdsl.core.renderer.Configuration
-import org.neo4j.cypherdsl.core.renderer.GeneralizedRenderer
-import org.neo4j.cypherdsl.core.renderer.Renderer
-import org.neo4j.cypherdsl.parser.CypherParser
-import org.neo4j.cypherdsl.parser.ExpressionCreatedEventType
-import org.neo4j.cypherdsl.parser.Options
 import org.neo4j.spark.cypher.CypherPreamble.fullPreamble
 import org.neo4j.spark.cypher.CypherRenderer
 import org.neo4j.spark.cypher.QueryEmbedder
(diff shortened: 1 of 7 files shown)


src/main/scala/org/neo4j/spark/util/Neo4jImplicits.scala
```diff
@@ -55,16 +61,49 @@ object Neo4jImplicits {
 
     def isQuoted(): Boolean = str.startsWith("`");
 
-    def removeAlias(): String = {
-      val splatString = str.unquote().split('.')
+    /**
+     * Splits a, possibly backtick quoted, property path into its segments.
+     *
+     * Dots inside backticks belong to the property name, so `` `table.key` `` is a single property
+     * named `table.key`, while `location.x` is the `x` field of the `location` property.
```
vasifabdulhamidov-neo, line 68: trying to understand: `table.key` is single property but `location.x` is not?  x means some nested field?

vasifabdulhamidov-neo (approved):
Lgtm 👍

src/main/scala/org/neo4j/spark/service/Neo4jQueryService.scala
```diff
@@ -412,16 +436,14 @@ class Neo4jQueryReadStrategy(
...
-        sourceNode
-      } else if (filter.isAttribute(Neo4jUtil.RELATIONSHIP_TARGET_ALIAS)) {
-        targetNode
-      } else if (filter.isAttribute(Neo4jUtil.RELATIONSHIP_ALIAS)) {
-        relationship
-      } else {
-        throw new IllegalArgumentException(s"Attribute '${filter.getAttribute.get}' is not valid")
-      }
+    def getAlias(filter: Filter): String = relationshipAliases
+      .find(filter.hasEntityAlias)
```
vasifabdulhamidov-neo, line 440: should we apply same check (used above for columnEntityAlias) `alias => name == alias || name == s"$alias>"` here ?

src/main/scala/org/neo4j/spark/util/Neo4jImplicits.scala
venikkin: In this example `location` is an alias name for a variable and `x` in a type-specific property of 2d/3d point: https://neo4j.com/docs/cypher-manual/current/values-and-types/spatial/. 
On the other hand: `table.key` is altogether a property name.

src/main/scala/org/neo4j/spark/service/Neo4jQueryService.scala
venikkin: `columnEntityAlias` is similar to `entityAlias`, but it also supports special `<..>` fields, like `<elementId>`, so it includes an extra check. `columnEntityAlias` is only used on `buildReturnExpression`. I can include it there so highlight it's specific for this context. Wdyt?

PR #1119, opened 2026-09-03
fix: security warnings wrt encryption settings

nord-erik:
Introduces two warnings:
- warn users who encrypt, but accepts all certificates
- warn users who connect without encryption

### DIFF
--- a/common/src/main/scala/org/neo4j/spark/util/Neo4jOptions.scala
+++ b/common/src/main/scala/org/neo4j/spark/util/Neo4jOptions.scala
@@ -507,7 +507,7 @@ case class Neo4jDriverOptions(
         if (!encryption) {
           logWarning(
             s"Connecting to ${primaryUrl.getScheme}://${primaryUrl.getHost} without any encryption! " +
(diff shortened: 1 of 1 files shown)


common/src/main/scala/org/neo4j/spark/util/Neo4jOptions.scala
```diff
@@ -505,14 +505,27 @@ case class Neo4jDriverOptions(
       case "neo4j+s" | "neo4j+ssc" | "bolt+s" | "bolt+ssc" => ()
       case _ => {
         if (!encryption) {
+          logWarning(
+            s"Connecting to ${primaryUrl.getScheme}://${primaryUrl.getHost} without any encryption! " +
+              s"Credentials will be sent in cleartext. To start encrypting, use a '+s' scheme (e.g. neo4j+s://) or set " +
```
fbiville: suggestion: I'd only recommend applying the URL scheme for this (I would not suggest using properties)

nord-erik: thanks, will update

venikkin (approved):

PR #1122, opened 2026-09-03

nord-erik:

identifier names.

Just using Cypher DSL v2022 creates regressions around two areas:
 - Sanitize values that start with "$" e.g. "$string"

Implemented in such way where these quirks are retained to keep
backwards compatibility.


vasifabdulhamidov-neo (approved):
Looks good!

PR #1123, opened 2026-09-03

nord-erik:

identifier names.

Just using Cypher DSL creates a regressions:

Implemented in such way where this regression is not introduced to keep
backwards compatibility.

### DIFF
--- a/src/main/scala/org/neo4j/spark/util/Neo4jImplicits.scala
+++ b/src/main/scala/org/neo4j/spark/util/Neo4jImplicits.scala
@@ -233,7 +233,7 @@ object Neo4jImplicits {
      * preserves that information once the reference is flattened into a single string.
      */
     def rawAttributeName(): String = {
-      predicate.references().head.fieldNames().map(_.quote()).mkString(".")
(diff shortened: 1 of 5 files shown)


src/main/scala/org/neo4j/spark/util/Neo4jImplicits.scala
```diff
@@ -52,14 +51,19 @@ object Neo4jImplicits {
...
+  implicit class CypherImplicits(str: String) {
 
-    def unquote(): String = str.replaceAll("`", "");
+    def sanitizeSchemaName(): String = {
+      if (str.startsWith(BACKTICK) && str.endsWith(BACKTICK)) {
+        SchemaNames.sanitize(str.substring(1, str.length - 1)).orElse("")
+      } else {
+        SchemaNames.sanitize(str).orElse("")
+      }
+    }
```
nord-erik, line 65: this whole bit here is to perserve some old edge cases so we're backwards compatible, yet safe

vasifabdulhamidov-neo (approved):

PR #1124, opened 2026-09-03
fix: resolve an ambiguity between alias separation and dots in property names for 5.0

venikkin:
This bug is related to all properties containing dot, not only packaged map structs.

The existing implementation treats dot primarily as separator between an alias and a property, and strips it when reading. For example flattened map property mymap.myproperty becomes myproperty and thus always read as null.

In reality, the dot has 3 possible meanings:

an alias
part of the property
nested field access (for example in points).
This PR addresses this ambiguity. It makes alias stripping explicit and requires the exact alias name, where possible. Filter encoding also uses extra backticks to distinguish between property name and field access.

Backported from https://github.com/neo4j/neo4j-spark-connector/pull/1117


nord-erik (approved):
indeed looks equivalent to 6.0 version, LGTM

PR #1144, opened 2026-09-08
feat: introduce an option to fail on unknown commit state

venikkin:
This PR introduces an option to decide if we can retry commits that failed in flight. The result of such commits is undefined and can result in duplicates.


vasifabdulhamidov-neo (approved):
💯

